PR #4, opened 2021-09-30
Remove hardcoded text

jarrodmillman:


layouts/partials/footer.html
```diff
@@ -1,12 +1,15 @@
+{{- $title          := .Site.Params.title }}
 {{- $footer      := .Site.Params.footer }}
 {{- $logo        := index $footer "logo" }}
 {{- $quickLinks  := index $footer "quicklinks" }}
 {{- $socialMedia := index $footer "socialmedia" }}
+{{- $navbarLogo     := .Site.Params.navbarlogo }}
```
stefanv, line 6: Shall we keep the capitalization constant throughout?

static/images/README.md
```diff
@@ -1,4 +1,19 @@
-inkscape -w 16 -h 16 -o 16.png logos/scipy.svg
-inkscape -w 32 -h 32 -o 32.png logos/scipy.svg
-inkscape -w 48 -h 48 -o 48.png logos/scipy.svg
+Add logo with name `logo.svg`.
```
stefanv, line 1: Looks like the main logo has a fixed filename, but the footer logo is still customizable?  Perhaps both or none shoul dbe.

stefanv (commented):
Looks good.  We may want to retain the camelCase for long parameter names in the configuration file?

layouts/partials/footer.html
jarrodmillman: I will do that in a followup PR, but I would first like to test this with numpy.org.

static/images/README.md
jarrodmillman: I think I got that one too, now.

PR #13, opened 2021-10-02
Add stubs for documentation

### LINKED ISSUE #3: Start documentation
Start README and docs.  For example,
- https://github.com/StefMa/hugo-fresh/tree/master/docs
- https://github.com/StefMa/hugo-fresh/blob/master/README.md

I can add some content later and it would be good to have a place to add docs when we add new functionality (e.g., tables).

jarrodmillman:
Close #3.


stefanv:
So, are we setting the repo as its own submodule? :)

I think the standard way to do this is to make a directory called `exampleSite` and to have that site use the current theme by setting the theme path to `../..`.  See, e.g., https://github.com/matcornic/hugo-theme-learn

Then, you can publish `exampleSite/public` via CircleCI / Netlifty etc.

PR #55, opened 2021-11-03
Tweak team CSS; use GH username if full name is missing

stefanv:
Addressing an issue @tupui and I noticed this morning


tupui:
Nice thanks!

PR #56, opened 2021-11-03
Add a shortcode to render and include markdown

stefanv:
@tupui Here's the promised shortcode.

I followed a [different tactic for scientific-python.org](https://github.com/scientific-python/scientific-python.org/pull/129/files#diff-383abe2a95758bff4e9ba060bc1ea9788b4527435c7144aa301b0dc4c7734e1dR1), but only because I was lazy and did not want to keep adding teams by hand.

In retrospect, this would have been simpler.

### DIFF
--- a/doc/content/example-markdown.md
+++ b/doc/content/example-markdown.md
@@ -1,0 +1,1 @@
+This is some *example markdown* with **bold**!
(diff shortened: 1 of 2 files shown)


tupui:
Great I will update my PR with this.

PR #66, opened 2021-11-09
Allow frontpage news links to redirect to arbitrary locations

stefanv:
/cc @tupui


tupui (approved):
Thanks @stefanv. I tested this locally and it works good and does not require much to update from the previous behaviour.

jarrodmillman (approved):

PR #67, opened 2021-11-09
Reorg docs

jarrodmillman:


doc/config.yaml
```diff
@@ -16,18 +16,22 @@ params:
   hero:
     title: Hugo Theme
     subtitle: Used by core projects in the scientific Python ecosystem.
-    buttontext: Go to docs
+    buttontext: Get started
```
tupui, line 19: I just had a stop when reading `Get started` at the top of the page and then `Getting started` as a section. Maybe we should rename one or the other.

doc/content/shortcodes.md
```diff
@@ -3,4 +3,115 @@ title: Shortcodes
 sidebar: true
 ---
```
tupui, line 5: I think we are missing a one liner to say what are these. When you are new to Hugo, it's not immediately clear where/how to use these.

tupui (approved):
LGTM, I have 2 minor suggestions.

doc/config.yaml
jarrodmillman: In b87e168, I changed ` ## Getting started ` to ` ## Download and install` in `doc/content/getstarted.md`.

doc/content/shortcodes.md
jarrodmillman: In b87e168, I added a short paragraph to the top of the file by editing `tools/render_shortcode_docs.py`.

@stefanv I also removed the old stub file for `doc/content/shortcodes.md`.  Instead I added some text in `tools/render_shortcode_docs.py` that will be added as a comment to the top of the  `doc/content/shortcodes.md` that says:
```
<!-- DO NOT EDIT.  This file is generated by `tools/render_shortcode_docs.py`.  -->
```

tupui: Perfect, thanks!

jarrodmillman:
@stefanv Would it be easy to add something like `shortcutDepth: 1` to the youtube shortcode?  Currently, the shortcodes documentation looks like:
screenshot

The `Open Data Science` entry shouldn't appear in the shortcuts list.

stefanv (approved):

stefanv:
> @stefanv Would it be easy to add something like `shortcutDepth: 1` to the youtube shortcode? Currently, the shortcodes documentation looks like: screenshot

Added in https://github.com/scientific-python/scientific-python-hugo-theme/pull/70

> Also, it looks like the order of the shortcut entries looks like it changes. We probably want them to always be in the same order. Alphabetical by shortcodename seems reasonable.

Currently it is organized by whatever order Python walks the path in.  Totally reasonable to add a `sorted(...)` around the filenames in `render_shortcode_docs.py`.

PR #72, opened 2021-11-10
Add linking to key features

### LINKED ISSUE #69: Feature Request: Make "keyfeatures" cards linkable
I think it would be a nice feature to allow the `keyfeatures` cards to link to other internal or external pages.

The hugo fresh theme has something similar: see the cards under the "Great Power Comes" section: https://hugo-fresh.vercel.app/ (though I think making the entire card clickable is preferable to adding a button; this is a minor detail though).

stefanv:
When a key feature has a link it will:

- Light up when hovering over it
- Have an animated line underneath it
- Have a clickable (hand) pointer

The other cards will be static.

Closes #69


rossbar (approved):
I tried this out in rossbar/manual.scientific-python.org-hugo#5 (haven't setup preview yet) and it works exactly as I'd hoped. Tested there with a relative URL to an internal page to compliment the example here with the external link - works as expected :rocket:

stefanv:
We don't have a use-case for this now, so closing until it comes up again.

jarrodmillman:
I have a potential use case, so want to check this out.  May want slightly different look / behavior.  Will add more later.

jarrodmillman:
See #147.

jarrodmillman:
See https://github.com/scientific-python/learn.scientific-python.org/pull/3.

Issue #10, opened 2021-10-01
Fade-in should be configurable


stefanv:
Currently, the sites fade in on load.  I don't like that behavior, and would like it to be configurable.

Depends on #9

PR #77, opened 2021-11-10
Make front page fading configurable

stefanv:
Will need to be re-enabled for `numpy.org` (and `scipy.org`, if they want it—@tupui).

Closes https://github.com/scientific-python/scientific-python-hugo-theme/issues/10


jarrodmillman (approved):

tupui:
I think for now we can leave it out for SciPy. To me the fading gives the impression that the website is slow to load.

Issue #1, opened 2021-09-29
Remove hard-coded bits


jarrodmillman:
Some of these can probably be left in, but it would be good to move some of this into the config file.  For example, the `Numpy logo` stuff.

```
$ grep -r 'NumPy'
layouts/partials/shell-lesson.html:>>> # The standard way to import NumPy:
static/css/styles.css:/* NumPy Color Palette */
static/gallery/team.html:    <title>NumPy</title>
static/gallery/team.html:          <h1 id="numpy-team-gallery">NumPy Team Gallery</h1>
```

```
$ grep -r 'Numpy'
layouts/partials/css.html:<!-- Numpy Favicon -->
layouts/partials/navbar.html:                <img src="{{ printf "/images/%s" $navbarLogo.image | relURL }}" alt="Numpy logo. A three-dimensional blue cube with the letter 'N' highlighted in yellow." width="112" height="28" alt="Numpy logo. A three-dimensional blue cube with the letter 'N' highlighted in yellow.">
layouts/partials/hero-body.html:                <img class="hero-logo" src="{{ printf "/images/%s" $image | relURL }}" alt="Numpy logo. A three-dimensional blue cube with the letter 'N' highlighted in yellow.">
```

```
$ grep -r 'numpy'
layouts/partials/footer.html:            <form action="https://numpy.us4.list-manage.com/subscribe/post?u=5ddd0d1d6e807900a8212481a&amp;id=287fa4253c" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate sign-up-container" target="_blank" novalidate>
layouts/partials/navbar.html:              <a href="https://crowdin.com/project/numpyorg"
layouts/partials/navbar-clone.html:              <a href="https://crowdin.com/project/numpyorg"
layouts/partials/shell.html:    <div class="numpy-shell-canvas">
layouts/partials/shell.html:        <div class="numpy-shell-container">
layouts/partials/shell.html:            <div class="real-shell" id="numpy-shell" style="display: none;">
layouts/partials/shell.html:                <pre id="demo-code" data-executable="true" data-language="python" style="white-space: pre-line;">import numpy as np</pre>
layouts/partials/shell-lesson.html:>>> import numpy as np
layouts/shortcodes/numfocus.html:    <a href="https://numfocus.org/donate-to-numpy">
[...]

PR #79, opened 2021-11-10
Render language dropdown in theme colors; move to partial

stefanv:
Uses the site parameter "crowdin" (URL) to link to the translation
effort.

Projects that want to render the translations differently can simply override the partial.

translations

The site parameter `crowdin` is undocumented.

### DIFF
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -108,6 +108,18 @@ em {
   font-weight: bold;
 }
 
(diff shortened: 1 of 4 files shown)


stefanv:
Partially addresses https://github.com/scientific-python/scientific-python-hugo-theme/issues/1#issuecomment-964828594

PR #80, opened 2021-11-15
Add team gallery

jarrodmillman:


doc/config.yaml
```diff
@@ -48,11 +48,13 @@ params:
       column1:
         title: ""
         links:
-          - text: Scientific Python Forum
-            link: https://discuss.scientific-python.org
-      column2:
-        links:
+          - text: About
+            link: /about/teams/
```
tupui, line 52: Looking above, I think this should just be
```suggestion
            link: /about/
```

stefanv (approved):

Issue #15, opened 2021-10-13
Analytics


jarrodmillman:
Consider alternative to google analytics.

PR #89, opened 2022-01-09
Add plausible analytics

stefanv:
Closes https://github.com/scientific-python/scientific-python-hugo-theme/issues/15

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -9,3 +9,6 @@ params:
   font:
     name: "Lato"
     sizes: [400,900]
+  plausible:
+    dataDomain: null
+    javaScript: "https://analytics.scientific-python.org/js/plausible.js"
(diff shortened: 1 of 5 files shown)


jarrodmillman:
@stefanv The CI should pass if you rebase on main.

tupui:
Considering all projects involved, it could almost make sense to build and host our own proxy (there are nodes with unlimited bandwidth available for less than $100/month on some platforms). Was there any discussion about that?

stefanv:
That's what we have; projects can just ask us, and we can add them.

stefanv:
@tupui See https://github.com/scientific-python/scientific-python-hugo-theme/pull/89/files#diff-d1b9763f2fd69c5bf0c8f0c2bd533c169f138d2f605b64a610a5a2fc59ec07d4R104

tupui:
> That's what we have; projects can just ask us, and we can add them.

Oh great! I will link this to the issue in SciPy then 😃

PR #92, opened 2022-01-23
Remove cruft

jarrodmillman:


stefanv:
NumPy uses this for their interactive prompt so we must just make sure it's in their custom JS.

PR #94, opened 2022-01-26
Remove uikit

jarrodmillman:
May still need this for numpy.org (need to check).  There are more issues to fix, so this isn't ready to merge.

- [ ] font spacing / weight
- [ ] paragraph spacing (not as much as before and more space below headers)
- [ ] feature (font / paragraph spaces?)
- [x] tt text (font weight / color)
- [x] border and background for code blocks

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -347,3 +347,9 @@ pre {
     border-radius: 3px;
     background: #fff;
 }
+
+code {
+    color: #f0506e;
+    padding: 2px 6px;
+    background: #f8f8f8;
+}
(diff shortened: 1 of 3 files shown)


stefanv:
Comparing:

https://deploy-preview-94--scientific-python-hugo-theme.netlify.app/
https://theme.scientific-python.org/

The only thing I notice is that text moves around a little bit, but more importantly that the footer icon colors change. That should be a small CSS adjustment to fix.

jarrodmillman:
screenshot

stefanv:
I'm not 100% sure what you're showing in that screenshot—this is not what I see on my screen (in yours, icons are missing and fonts seem to be in the same places).

jarrodmillman:
There is also a spacing issue (e.g., codeblocks) that @stefanv noticed.

PR #95, opened 2022-01-26
Remove jquery

jarrodmillman:
I think this should be good to go.


tupui (approved):
Will definitely help to speed things 👍 if we don't need it later then perfect.

stefanv:
I'll take a look at how to fix the shortcuts.

stefanv:
Jquery is being used in fresh.js and shortcuts.js quite extensively, so it will take a bit more work to get rid of it.

stefanv:
The simplest solution for now may be to replace jquery with its slim build, https://code.jquery.com/jquery-3.6.0.slim.min.js, which is 70KB.

PR #99, opened 2022-01-27
Add lighthouse in CI

tupui:
Based on https://github.com/pydata/pydata-sphinx-theme/blob/a2497c60b815d4a09ead8ac154aa6d08e7a9ab50/.github/workflows/tests.yml#L75-L148

We could do more involved stuff https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md but I think it's a bit overkill.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
         uses: peaceiris/actions-hugo@v2
         with:
           hugo-version: 'latest'
-          # extended: true
+          extended: true
 
       - name: Build
         run:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,9 @@ jobs:
           # extended: true
 
       - name: Build
-        run: cd doc && hugo --minify
+        run:
+          python tools/render_shortcode_docs.py > doc/content/shortcodes.md
+          cd doc && hugo --themesDir="../.." --minify
 
       # Serve the docs and wait to be ready
       - name: Serve the built site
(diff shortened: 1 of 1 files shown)


stefanv:
Neat!!

How hard would it be to surface the lighthouse report as a hosted artifact?

Should we also be using --minify on our other hugo builds?

tupui:
> Neat!!
> 
> How hard would it be to surface the lighthouse report as a hosted artifact?

I removed this part but if you go to the pydata theme they store the artifact. Is this what you are talking about? Or you want to download it and put it somewhere for us to see the evolution over time?

> Should we also be using --minify on our other hugo builds?

I think so, just seen this on Hugo's website when doing this.

tupui:
This
```yaml
      - name: Publish Audit reports
        uses: actions/upload-artifact@v2
        with:
          name: Pa11y and Lighthouse ${{ github.run_number }}
          path: |
            /tmp/pa11y
            /tmp/lighthouse
        if: always()
```

tupui:
Thanks @rossbar! Is it working good then? You can share an URL?

jarrodmillman:
I am not sure we should be running lighthouse w/ GH actions.  I am going to try something and will let you know if it works later today.

rossbar:
> Thanks @rossbar! Is it working good then? You can share an URL?

I got it to the point where the build was passing and lighthouse ran in the action, but something is wrong with accessing the hosted site in the job because the scores for everything are coming back as `0`, e.g. : https://github.com/rossbar/scientific-python-hugo-theme/runs/4972834772?check_suite_focus=true

I think @jarrodmillman found a way to run lighthouse within netlify, so I'm going to hold off on looking at the action while he tries that.

jarrodmillman:
See #101.

PR #107, opened 2022-02-01
Do not use FontAwesome for scroll-up icon

stefanv:
Closes https://github.com/numpy/numpy.org/issues/544


rossbar (approved):
I tried this building the numpy.org site locally and it works as expected :+1:

PR #110, opened 2022-02-01
Remove hero-footer

jarrodmillman:
We use `layouts/partials/footer.html`.  I am pretty certain it is safe to remove this, but I would be happy for someone else to double check that I didnt' overlook anything.


tupui (approved):
+1 on removing this. At least SciPy and NumPy don't use it and I don't really see a use case for them. Could be used though on the site of the theme to show other projects as "clients". But a gallery like with links would be better (because right now this footer does not support links)

PR #111, opened 2022-02-01
Use /js/script.js instead of /js/plausible.js for analytics

stefanv:
Many ad-blockers won't allow plausible.js through unless it is run on
plausible.io itself.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -8,4 +8,4 @@ params:
     sizes: [400,900]
   plausible:
     dataDomain: null
-    javaScript: "https://analytics.scientific-python.org/js/plausible.js"
+    javaScript: "https://analytics.scientific-python.org/js/script.js"


tupui:
Oh so this one means we need to update SciPy.org to the latest version of the theme right?

stefanv:
The old script tag still works, but this one will work better.

A bigger problem right now is that ublock origin stops `analytics.*/.*event` from going through, which means we cannot track clicks anyway.  So, we're thinking of renaming the domain from analytics.scientific-python.org to ?.scientific-python.org.  Any suggestions for `?`?

tupui:
So we need to hide ourselves. A few things come to mind:

* `see`, `vision`, `clarity`, `reveal`
* `studio`, `stats`, `probe`, `sense`, `sentinel`
* `idea`, `hint`
* `roots` (like we have a tree as our logo, roots to link to us, I am soon a poet 😅)

stefanv:
:) What about `views.scientific-python.org`?

stefanv:
@tupui The rename is complete, so please update SciPy accordingly; thank you!

PR #115, opened 2022-02-02
Update title banner

jarrodmillman:


stefanv:
This changes the way the theme docs renders, if I'm not mistaken? I.e., there's no more example now of the getting started button feature.

PR #116, opened 2022-02-03
Highlight current heading in shortcuts

stefanv:
Also:

- Add a list icon a-la pydata-sphinx-theme (but use MUI icon instead of FontAwesome)
- Improve the way the shortcut list is generated: anchors now come directly from Hugo
- Prevent shortcut list from scrolling off-screen


jarrodmillman:
Did you want to change the name from `shortcuts` to `sidebar` or something else?

stefanv:
I was thinking we can do that in another PR.

jarrodmillman:
LGTM!!

PR #120, opened 2022-02-15
Add mermaid shortcode

jarrodmillman:
@stefanv I am having trouble figuring out why the example text is being rendered:
screenshot

- [ ] The example is rendered, but it is supposed to show the text.
- [ ] The text is also getting clipped in the rendered version.  I don't know when this started, but it used to not do this.  I know it worked correctly before we switched to scientific-python-hugo-theme, but I don't remember if it ever rendered correctly after we switched.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ teams-clean:
 
 teams: | teams-clean $(patsubst %,$(TEAMS_DIR)/%.md,$(TEAMS))
 
-doc/content/shortcodes.md:
+doc/content/shortcodes.md: $(wildcard layouts/shortcodes/*.html)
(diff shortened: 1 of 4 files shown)


tupui (commented):
Seems like it's rendering twice instead of showing the raw input. Also the items do not fit in the boxes.

tupui:
Screenshot 2022-02-15 at 20 35 28

jarrodmillman:
The example is no longer rendered after adding a space around `mermaid` and `\mermaid`.  The rendered version still cuts off the text.
screenshot

If, under `Inherited from g`, I unselect `font-weight: 700;` under `.label` provided by bulma.min.css, then it looks better:
after

PR #124, opened 2022-02-25
Switch to a FontAwesome 6 kit

stefanv:
A kit is a personalized FA download. You need a FontAwesome account to
make one, but it's free: https://fontawesome.com/v6/docs/web/setup/use-kit

In order to accommodate all FA icons, icon names are now specified differently:

`fa-brands fa-github` instead of `github`

These names can be found at https://fontawesome.com/search

The kit download should be significantly smaller than what we had
before, with icons being downloaded on-demand.

### DIFF
--- a/doc/config.yaml
+++ b/doc/config.yaml
@@ -42,7 +42,8 @@ params:
     socialmediatitle: ""
     socialmedia:
       - link: https://github.com/scientific-python/
-        icon: github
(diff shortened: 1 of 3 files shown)


jarrodmillman:
We should add this to the documentation.

jarrodmillman (approved):

PR #131, opened 2022-03-03
Set width and height for bulleted list icon

### LINKED ISSUE #130: shortcuts title / list-bulleted issue
It looks like there is a minor issue (maybe it happened when `mui-icons_format-list-bulleted` to `list-bulleted`.
screenshot

Previously, the text `On this page` appeared directly beside the bulleted list icon.

jarrodmillman:
Fixes #130.  Should we set the `width`, `height`, or `viewBox` on all the icons?  Currently some have it and some don't.  

For example, [assets/icons/discourse.svg](https://github.com/scientific-python/scientific-python-hugo-theme/pull/129/files#diff-321e9619c93785533387121208fdc1e752f23b2d60d9eb48618c8f42b2db0ffa) has `viewBox="0 0 24 24"`, but [assets/icons/facebook.svg](https://github.com/scientific-python/scientific-python-hugo-theme/pull/129/files#diff-8cb87ee9035afec71ef10e3f0f596ef838d1ecb0c92f6a80895a5e2447f1f565) doesn't.  I am not sure what the best practice here is.


tupui (approved):
If that fixes it then maybe we should add it everywhere as a safeguard. I suppose users can override this while using it right?

stefanv:
Let me rescale the icon by hand to be of the correct size.  All the other icons already have this scaling built in.

stefanv:
See https://github.com/scientific-python/scientific-python-hugo-theme/pull/133

PR #133, opened 2022-03-04
Update list icon

stefanv:
See #131


jarrodmillman:
I would like to understand why this is a better solution than #131.  Let's discuss this at some point today.

PR #147, opened 2022-03-09
Add front page panel

jarrodmillman:
This is not ready to merge or review.  I am in the process of adapting the keyfeatures style to a panel style.  I thought about starting with the fresh feature-card thing, but I thought this would be simpler.  For now, you can see a screenshot of what this roughly looks like here:  https://github.com/scientific-python/learn.scientific-python.org/pull/3


jarrodmillman:
Decided to work on this in the learn.scientific-python.org repo.

PR #151, opened 2022-03-12
Use dashes instead of underscores for consistency

stefanv:


stefanv:
I don't think this is used elsewhere, so I am going to merge.

PR #157, opened 2022-03-16
Add proof-of-concept dark mode switcher

stefanv:
darkmode


tupui:
Good start. Switching is fast. Now I suppose the hardest part is to find a colour scheme which offers good readability for code/text/elements. I am very bad at this I am afraid, but maybe @MarsBarLee would be interested in this?

PR #159, opened 2022-03-21
Add pre-commit hook / linter

jarrodmillman:


tupui:
haha I see what you did here 😉

.github/workflows/lint.yml
```diff
@@ -0,0 +1,26 @@
+name: style
+
+on: [push, pull_request]
+
+jobs:
```
tupui, line 5: Why not using pre-commit there too?

```
jobs:
  pre-commit:
    name: Linting
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-python@v2
    - uses: pre-commit/action@v2.0.0
```

PR #163, opened 2022-03-22
Proof of concept of SVG icons next to external header links

stefanv:
Will necessitate updating blog navigation partial as well (I should fix that in the theme, so that the blog can re-use the navbar from here).

- [ ] Remove external_link from Features; it's just to show how it works


jarrodmillman (approved):

jarrodmillman:
I am merging this with the example external link for now.  I will make an issue and follow-up PR to remove it and document the feature.

PR #167, opened 2022-03-24
Remove panel underline of button

jarrodmillman:


tupui (commented):
Actually we have to change the css variable instead. (and also I tested this locally, we should add a margin 15px to `.panel-underline`)

jarrodmillman:
@tupui This should work.

tupui:
As discussed, I will take over this PR.

assets/theme-css/panel.css
```diff
@@ -30,8 +30,9 @@
     height: 140px;
     min-width: 275px;
     width: 335px;
-    margin: 25px auto;
+    margin: 25px 15px 25px 15px;
     border-radius: 3px;
+    box-shadow: 0px 8px 15px var(--colorShadow);
```
jarrodmillman, line 35: I added it when I removed the panel-underline thing.  Now that it is back, I don't think this is needed.

jarrodmillman:
I tested this on both learn and blog and it LGTM!  I made one comment regarding removing one line that I don't think is necessary.

assets/theme-css/panel.css
tupui, line 35:
```suggestion
```

tupui: Done

Issue #170, opened 2022-04-21
Hamburguer menu for top bar not working


melissawm:
When the browser window size is reduced, or the page it viewed on mobile, there is a hamburguer menu that should show the top navbar items when clicked on. This is currently not working, as the hamburguer menu seems to be empty.

Example: go to numpy.org, reduce the window size until the hamburguer appears, and click it. No menu appears, but we do see the hamburguer change to an X indicating this menu can be closed.

PR #172, opened 2022-04-26
Add back id for navbar-menu so hamburger folding works again

stefanv:
Closes gh-170


melissawm (approved):
Works great - thanks @stefanv !

jarrodmillman (approved):

PR #180, opened 2022-05-12
Reduce fresh dependency

jarrodmillman:
Trying to get a sense of how much of the fresh theme we are actually using.


jarrodmillman:
I've manually tested this on the various websites using this theme and everything looks fine to me. I am going to merge this and create draft PRs on the various websites. If we run into any issues with the draft PRs, I will fix the theme or revert this PR.

PR #182, opened 2022-06-24
Center footer elements

tupui:
When using small screens like a phone, the elements of the footer are not centred. This fixes it.


tupui:
Also seeing that there is no padding on the text+logo if text is too long?? But not related to this.

jarrodmillman (approved):

PR #191, opened 2022-08-08
Add twitter cards

jarrodmillman:
See https://github.com/numpy/numpy.org/pull/588.


jarrodmillman:
I am not sure if this worked. Before
screenshot
 
Clearly, no card found.  Now, we get
screenshot

It says the card was loaded successfully, but it is still "Unable to render Card preview".

PR #199, opened 2022-09-09
Use jquery.slim

jarrodmillman:


stefanv:
I thought we did this before and ran into missing functionality. Maybe that functionality has been removed meanwhile?  I think it had to do with mouse clicks, touchpad, or menu bar, but my memory is fuzzy.

jarrodmillman:
There is this #106, but I couldn't find the bug report. @stefanv Do you know where the animation functionality is used?

stefanv:
Nice find. That issue says we need the animation functions; so perhaps just see if those are still utilized anywhere (menu bar rollout?). We can also try opening the web debug dialog to see if JS functions are called but missing.

stefanv:
I took a look; I think it's only the fading functions in `fresh.js`.

If we get rid of the preloader, then we should be good (but I suppose numpy and some others still use it?).

alphapapa:
@stefanv @jarrodmillman Would you like me to look into whether we should still try to do this?

PR #218, opened 2023-03-02
MAINT: rename icons section to avoid adblock filtering

bsipocz:
I couldn't fully test this locally as I was getting a make error, but I am almost certain this would fix the issue I see, namely that I don't see the community and social media icons in the footer.

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -292,7 +292,7 @@ blockquote p {
     max-width: 50vw;
   }
 
-  .social-media-icons {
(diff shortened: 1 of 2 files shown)


bsipocz:
Indeed, I see them in the rendering above, while I don't see them in the current deployment at https://theme.scientific-python.org/ . 🎉 

Screenshot 2023-03-01 at 22 54 57

stefanv (approved):

bsipocz:
I go ahead and merge, so I'll have the fix :)

PR #228, opened 2023-04-25
Update mermaid

jarrodmillman:
Close #228.

Mermaid 10 is ESM only, see
https://github.com/mermaid-js/mermaid/blob/master/CHANGELOG.md

Looks reasonable. Compare:
https://deploy-preview-228--scientific-python-hugo-theme.netlify.app/shortcodes/#mermaid
https://theme.scientific-python.org/shortcodes/#mermaid

Doesn't seem to have hurt our metrics:
https://deploy-preview-228--scientific-python-hugo-theme.netlify.app/reports/lighthouse
https://theme.scientific-python.org/reports/lighthouse

### DIFF
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -35,5 +35,8 @@
 {{ end -}}
 
 {{- if .HasShortcode "mermaid" }}
-<script async src="https://unpkg.com/mermaid@9.4.3/dist/mermaid.min.js"></script>
+<script type="module">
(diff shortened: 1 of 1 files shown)


jarrodmillman:
I wasn't familiar with ESM, so I wasn't sure if we should merge this or not. Do we need to worry about legacy browser issues?

stefanv:
ESM has been supported since 2018 in all major browsers, so I think we're fine.

PR #230, opened 2023-05-09
Fix: (layouts/partials/footer.html) Alignment

### LINKED ISSUE #192: Alignment issues
I have noticed a few alignment issues on the footer and hero banner. Interestingly it seems fine on SciPy's page for instance. Maybe this just happens if elements are too big.

The footer is not centered if you resize the window to a small screen. Some elements do not wrap and just disappear. Lastly, the vertical alignment is not correct and I think we should reduce the footer size.
Screenshot 2022-08-05 at 19 42 02

There is no padding on the hero banner when resized
Screenshot 2022-08-05 at 19 42 40

alphapapa:
This removes the hard-coded padding-left property from these UL elements.  It fixes the alignment in a narrow (e.g. smartphone) display, and in a wide (e.g. desktop) display, the difference is insignificant and no worse.

Fixes #192.  Thanks to Pamphile Roy (@tupui) for reporting.

#### Before changes

before-narrow
before-wide

#### After changes

after-narrow
after-wide

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -52,11 +52,12 @@ em {
   font-weight: 900;
   font-size: 6em;
   color: var(--colorPrimaryDark);
+  margin: 0.5em;
 }
 
 .hero-logo {
   max-height: 75px;
-  padding: 0px 0 0 15px;
+  margin: 1em;
 }
 
 .hero-subtitle {
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -182,21 +182,11 @@ p {
   .hero-title {
     font-size: 4em;
   }
-
-  .hero-logo {
-    max-height: 60px;
-  }
 }
 
 @media only screen and (max-width: 400px) {
-  .hero-logo {
-    max-height: 50px;
-  }
-}
-
-@media only screen and (max-width: 320px) {
-  .hero-logo {
-    max-height: 40px;
+  .hero-title {
+    flex-direction: column;
   }
 }
 
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -33,12 +33,11 @@ em {
 }
 
 .hero-title-content {
+  align-items: center;
   display: flex;
   flex-direction: row;
-  width: 100%;
-  max-width: 1200px;
-  margin: 6vh auto;
   justify-content: space-around;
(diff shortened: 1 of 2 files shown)


alphapapa:
The second commit adjusts margins around the `.hero-title` and `.hero-logo`.  I'm not sure that this is ideal yet, but it prevents squashing elements too close together or too close to the screen edge.

A noticeable difference is that the "Scientific Python" text is more often split onto separate lines, depending on screen size.  At some sizes, this is inevitable, but at wider ones it could be avoided by widening the containing element.  But some of this is a matter of style or preference.

@jarrodmillman @tupui What do you think?  Thanks.

# Screenshots

## Before:

hero-title-before-wide
hero-title-before-narrow

## After:

hero-title-after-wide
hero-title-after-narrow

alphapapa:
Hi Pamphile,

> For the first commit, it's way better! There still seems to be a slight padding, can this be removed??

I'm not sure what padding you mean.  Would you be more specific please?

> For the second commit. I think it looked better with the logo close to the title. Otherwise we have a lot of empty space in between. But in general I do agree that this could be re-designed.

I agree that the spacing seems excessive.  I'll try to improve that and then make another commit on this PR.

Thanks for your feedback.

alphapapa:
I've made some further improvements to the CSS.  Here are screenshots showing phone, tablet, and desktop screens:


phone
tablet
desktop

I also improved the alignment of the "GET STARTED" button and the text above it.

This is much better, I think.  The only minor problem I still see is that, in the phone-sized screenshot (the first one), the logo is very small.  This could probably be further improved, and the best way to do so would probably be--on narrow screens, specifically--to either put the logo below the text rather than beside it, or decrease the text size to allow more room for the logo.

What do you all think?  Thanks.

stefanv:
I agree with your assessment about the logo on mobile, and the solution proposed (top, perhaps?) sounds good.

alphapapa:
@stefanv Thanks.  I made some more adjustments and I think it looks pretty good now.  Here are screenshots at relevant sizes:

scientific-python org_
scientific-python org_ (1)
scientific-python org_ (2)
scientific-python org_ (3)

alphapapa:
When approved, this should probably be squash-merged.  Thanks.

jarrodmillman (approved):

PR #235, opened 2023-05-31
Meta: Add .dir-locals.el

alphapapa:
To ensure Emacs indents CSS files consistently with the project's existing 2-spaces-per-level.  (This file will likely be useful for other file types as well.)


stefanv (approved):
The emacs default is 4, so this makes sense.

PR #238, opened 2023-06-18
Change the width of the cta-button

InessaPawson:
Changing the width of the cta-button. 
Refer to the attached image to see how it will look after the change.

numpyorg-ctabutton-440px


stefanv:
Using rem or em units may be better for scaling.

jarrodmillman (approved):

PR #245, opened 2023-06-27
Revert "Change the width of the cta-button (#238)"

jarrodmillman:
This reverts commit fefe6d0e9e5bcdc28b3a33c462e93db9f1633b1a. This made the button too big for other projects. So we will make these changes to numpy.org directly instead.


stefanv:
See also #246

PR #249, opened 2023-07-20
Allow navbarlogo to always show

jarrodmillman:
On blog.scientific-python.org and learn.scientific-python.org, we don't want to display the hero.html partial and instead want to always show the navbarlogo.


brianhawthorne (approved):
LGTM

PR #257, opened 2023-09-14
Update hugo

jarrodmillman:


doc/netlify.toml
```diff
@@ -1,24 +1,9 @@
...
   PYTHON_VERSION = "3.8" # netlify currently only support 2.7 and 3.8
-  HUGO_VERSION = "0.116.1"
+  HUGO_VERSION = "0.118.2"
 
 [context.deploy-preview]
   ignore = "false"
 
 [[plugins]]
   package = "netlify-plugin-checklinks"
-
```
alphapapa, line 10: Just curious, is that plugin built-in to Hugo 0.118, or...?

alphapapa (approved):

doc/netlify.toml
jarrodmillman: I took that out because you can enable it in the netlify UI. If you do it in the config file it won't update to a new version, but if you use the UI it automatically updates.

alphapapa: Got it, thanks.

Issue #261, opened 2023-09-15
Restore project icon positioning


stefanv:
In https://github.com/scientific-python/scientific-python-hugo-theme/pull/230, I didn't notice how much the logo size and positioning changed. It would be good to get them back to the way they were:

<img src="https://user-images.githubusercontent.com/601365/236992421-69a6ffc5-28f7-4b3f-ba3a-39a6d2da9068.png"/>

It's especially noticeable on http://scipy.org

jarrodmillman:
See https://github.com/numpy/numpy.org/pull/668#issuecomment-1721771418

jarrodmillman:
I am not sure I like the alignment we had before. I think it looks a little off that Community developed and the Get Started button appear to be left aligned. But I don't have a strong opinion.

PR #265, opened 2023-09-26
Fix: Project icon positioning (#261)

alphapapa:
This moves the project icon to appear like a superscript.  See #261.

Screen Shot 2023-09-25 at 22 00 16

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -56,12 +56,15 @@ em {
   font-size: 6em;
   color: var(--colorPrimaryDark);
   justify-content: center;
+  margin-bottom: 0.2em;
   margin-right: 0.5em;
 }
 
 .hero-logo {
-  max-height: 2em;
-  padding: 0.25em;
+  align-self: start;
+  margin: 0.1em;
+  margin-left: 0.25em;
+  max-height: 0.8em;
 }
 
 .hero-subtitle {


stefanv:
It might be my imagination, but that icon looks higher than it used to be. Could you please compare against an older version of the theme?

jarrodmillman:
@stefanv It is definitely higher and smaller. Do you prefer it lower and bigger? Or just lower, but this size?

stefanv:
I would like it to be restored to how it was, then we can discuss modifications another time.

alphapapa:
> I would like it to be restored to how it was, then we can discuss modifications another time.

If we revert that, it would also mean reverting the fix that was merged in https://github.com/scientific-python/scientific-python-hugo-theme/pull/230, and having to redesign that somehow.

I just pushed another commit to this PR that looks like this:

Screen Shot 2023-09-26 at 12 14 59

I tried to imitate the screenshot in https://github.com/scientific-python/scientific-python-hugo-theme/issues/261#issue-1899012424 as closely as I could.  It's mostly just a matter of tweaking the `max-height` on the `.hero-icon`.

stefanv:
Thanks, Adam. I think if we add a tiny bit of extra space between the text and the logo, that's close enough and we can merge.

alphapapa:
> Thanks, Adam. I think if we add a tiny bit of extra space between the text and the logo, that's close enough and we can merge.

Sure, just pushed another commit that looks like this:

localhost_1313_

stefanv:
Worth checking different zoom levels, and also the phone/tablet emulation in the web debugger window.

jarrodmillman:
Here is what the currently deployed website looks like.
2023-09-26T12:07:24,756360960-07:00

jarrodmillman:
Here is what it looked like 5 months ago (e3516b2):
2023-09-26T12:14:48,927545478-07:00

stefanv:
Looks like it's been present, in which case we can merge this and file a new issue for it.

jarrodmillman (approved):

alphapapa:
> Looks like it's been present, in which case we can merge this and file a new issue for it.

Will do.  Thanks.

PR #266, opened 2023-09-26
Update hugo

jarrodmillman:


jarrodmillman:
The error was a netlify issue.

Issue #256, opened 2023-08-24
Notice boxes' CIRCLED INFORMATION SOURCE character may appear as tofu


alphapapa:
@jarrodmillman showed me these screenshots of how the notice boxes on the https://scientific-python.org/about/ page look on his Fedora system:

![tofu1](https://github.com/scientific-python/scientific-python-hugo-theme/assets/601365/64bd1030-cde1-4ae2-b880-af90dfdda227)
![tofu2](https://github.com/scientific-python/scientific-python-hugo-theme/assets/601365/c9698aa8-3823-4c9a-8680-14e05f138cff)

We may be able to improve this by ensuring that we select a fallback font that includes this character.

stefanv:
It renders correctly on my Fedora system, so I suppose it's a font issue indeed. Perhaps we should set more stringent font requirements and fallbacks specifically on the class attached to the icon, instead of to the whole website (I'm not sure those fallback routes will make sense to have the same, but could be).

PR #268, opened 2023-09-27
Add: Fallback font for symbols

alphapapa:
See <https://github.com/scientific-python/scientific-python-hugo-theme/issues/256>.

### DIFF
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -15,11 +15,7 @@
 <link rel="stylesheet" href="{{ $fontUrl }}">
 
 <!-- Fallback font for symbols (such as "🛈"). -->
-{{- $fallbackFontName     := .Site.Params.font.fallbackName | default "Noto Sans Symbols 2" -}}
-{{- $fallbackFontFace     := replace $fallbackFontName " " "+" -}}
-{{- $fallbackFontSizes    := delimit (.Site.Params.font.sizes | default (slice 300 400 600 700)) "," -}}
-{{- $fallbackFontUrl      := printf "https://fonts.googleapis.com/css?family=%s:%s" $fallbackFontFace $fallbackFontSizes -}}
-<link rel="stylesheet" href="{{ $fallbackFontUrl }}">
+<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Noto+Sans+Symbols+2">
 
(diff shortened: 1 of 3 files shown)


alphapapa:
@jarrodmillman This fallback font, *Noto Sans Symbols 2*, includes the `🛈` character, so this should fix rendering on your system and any other that doesn't have a system font which includes the character.  Please let me know if this deploy preview renders it correctly on your system: https://deploy-preview-268--scientific-python-hugo-theme.netlify.app/shortcodes/

jarrodmillman (approved):
LGTM

layouts/partials/css.html
```diff
@@ -13,6 +13,14 @@
 
 <link rel="icon" href="{{ "images/favicon.ico" | relURL }}" />
 <link rel="stylesheet" href="{{ $fontUrl }}">
+
+<!-- Fallback font for symbols (such as "🛈"). -->
```
stefanv, line 17: Why don't we just add this as a static entry, instead of making it configurable?

I also don't think we need 300, 400, 600, and 700 just for the symbol.

alphapapa: Changed in https://github.com/scientific-python/scientific-python-hugo-theme/pull/268/commits/2e0689fb160a56b7e6dc1a86d65d2de63e116c38.  Thanks.

jarrodmillman:
hmm. It worked for me before, but I am now getting
2023-09-27T10:43:12,405714269-07:00

stefanv:
That header has font weight 700, I think, so we'd need that weight included. Perhaps 400, 700 to have bold and non-bold.

jarrodmillman:
2023-09-27T10:59:37,939673052-07:00

alphapapa:
> That header has font weight 700, I think, so we'd need that weight included. Perhaps 400, 700 to have bold and non-bold.

Unfortunately, the *Noto Sans Symbols 2* font only offers one weight, 400: https://fonts.google.com/noto/specimen/Noto+Sans+Symbols+2  If you'd like, I could try to dig up another fallback font that includes that symbol, but I'm not sure if there's much alternative.

This latest screenshot from Jarrod shows how it looks on his system now.  What do you think?

stefanv:
That looks great!

stefanv (approved):

PR #293, opened 2023-10-17
Meta: (Makefile) Serve with --poll, etc.

alphapapa:
Using --poll seems to help work around
<https://github.com/gohugoio/hugo/issues/10893>.


jarrodmillman (approved):
The CI failure is unrelated. We should look into why it keeps happening at some point.

PR #301, opened 2023-10-19
Use pydata-sphinx-theme admonition styling

jarrodmillman:


alphapapa:
@jarrodmillman Would you like me to continue working on this?  e.g. we could remove the Font Awesome icons, or we could add Font Awesome to the theme, like in pydata-sphinx-theme, so those icons could be used to replace the information circle character.

PR #303, opened 2023-10-20
Use dart sass

jarrodmillman:
See https://sass-lang.com/blog/libsass-is-deprecated/ and https://gohugo.io/hugo-pipes/transpile-sass-to-css/#dart-sass.

We will need to update the various sites and then we can make a new PR to change this line in the netlify.toml file.
```
    (cd .. ; make docs) \
```
to
```
    (cd .. ; make theme) \
```

We will also need to add documentation for the INSTALL, but we can do that in the PR that fixes the netlify.toml file.


jarrodmillman:
I am not sure why, but the lighthouse performance took a big hit with this PR. It was 96 now it is 87.

jarrodmillman:
It was 95 with this PR https://github.com/scientific-python/scientific-python-hugo-theme/pull/301.

Issue #299, opened 2023-10-18
Multisite testing follow-up


jarrodmillman:
I am not sure what the best way to test everything is, but it would be nice to see how changes to the theme impact multiple sites (during PR review in particular). So I first made it easier to control how netlify built the site in #297 and then added extra sites to the deploy preview in #298.

There is no link or documentation now, but for any PR preview like:
https://deploy-preview-298--scientific-python-hugo-theme.netlify.app/

you can now add scipy, blog, or learn and you will get something like:
https://deploy-preview-298--scientific-python-hugo-theme.netlify.app/scipy/
https://deploy-preview-298--scientific-python-hugo-theme.netlify.app/blog/
https://deploy-preview-298--scientific-python-hugo-theme.netlify.app/learn/

jarrodmillman:
We should review this carefully. I am sure there will be issues.

- [ ] how do we handle previews vs. actual deployment (it would be nice if https://theme.scientific-python.org/scipy doesn't exist.

tupui:
Oh this is really convenient 👍👏

stefanv:
Closing this for now, as we are only building other sites during PR previews.

PR #306, opened 2023-10-23
Only build site previews for PRs

stefanv:
See #299.


jarrodmillman (approved):

jarrodmillman:
Not sure why, but the lighthouse performance score dropped 10 points.

PR #312, opened 2023-10-24
Add missing admonitions

jarrodmillman:


layouts/shortcodes/admonition.html
```diff
@@ -2,26 +2,50 @@
...
-This is a tip.
+{{< admonition warning>}}
+Reader discretion is strongly advised.
 {{< /admonition >}}
 
 */}}
 
 <div class="admonition {{ .Get 0 }}">
-  <div class="admonition-title">🛈  &nbsp;{{ .Get 0 | humanize }}</div>
+  <div class="admonition-title"> &nbsp;{{ .Get 0 | humanize }}</div>
```
stefanv, line 48: Remove either the &nbsp or the space; I think with both we are wider than PDST.

PR #320, opened 2023-10-25
Purge unused content-container

jarrodmillman:


stefanv:
See https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/layouts/partials/single/content.html and customizations on the blog for examples of where content-container is used.

PR #331, opened 2023-10-26
Remove jquery usage in theme JS

### LINKED ISSUE #222: Remove jquery
It is likely that we don't use much of it, but should investigate.

stefanv:
Another attempt to banish jquery! :sweat_smile: 

Closes #222

### DIFF
--- a/assets/js/fresh.js
+++ b/assets/js/fresh.js
@@ -37,50 +37,6 @@ function whenReady() {
     );
     return false;
   });
-
-  // Select all links with hashes
-  $('a[href*="#"]')
-    // Remove links that don't actually link to anything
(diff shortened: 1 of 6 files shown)


jarrodmillman:
I was thinking of using the hamburger menu from pydata-sphinx-theme.

stefanv:
Do they use jquery? If not, that'd be good, but perhaps a bigger operation than what I have in mind here (just straightforward translation of jquery to vanilla JS).

jarrodmillman:
I haven't looked into itmuch, so not sure it is even viable. Just thought I'd mention it.

jarrodmillman (approved):

PR #338, opened 2023-10-30
Render mermaid via code blocks, instead of shortcodes

stefanv:
Will require a change to SPECs.


jarrodmillman (approved):
The failure is expected as we need to update the SPEC repo when we update the theme.

PR #347, opened 2023-10-31
Remove bulma subheading styling

jarrodmillman:
See the subheading at the top of https://scientific-python.org/summits/developer/2023/ in dark mode.


stefanv:
I think it would be good to keep *some* styling for the subtitle, since PDST doesn't have subtitles. The following in `styles.css` should suffice:

```css
.subtitle {
  font-style: italic;
  margin-top: -0.5rem;
}
```

stefanv (approved):

PR #348, opened 2023-11-01
Upstream blog formatting

stefanv:
This can be used along with https://github.com/stefanv/blog.scientific-python.org/tree/use-upstream-blog-layout

That blog checkout has to be updated to use *this theme* however; I just symlinked the theme.


stefanv:
In a follow-up PR, I will try to better align the containers we use with PDST.

PR #349, opened 2023-11-01
Add pydata-sphinx-theme blog example

jarrodmillman:


stefanv:
Need to:

1. Add `layout` or `type` metadata to blog post that is outside of either `/posts` or `/blog`.
2. Use correct metadata fields (ours are a bit different from theirs, in that we use `authors`, `tags` is a list, and we only use tags (no categories).
3. Document how to add `/blog` / `/posts` (easy, just put posts in there, as well as `_index.md` files for each subdirectory---see blog.scientific-python.org as an example), or elsewhere (harder, need to use the `layout` or `type` metadata).

stefanv:
```
layout: posts/single
```

stefanv:
Also document that:

```
disableKinds: ["term", "taxonomy"]
```

In `config.yaml` would prevent tags from rendering; ensure `term` is not in there.

doc/content/examples/blog/_index.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Blog
+type: posts
```
stefanv, line 3: We may want to allow `post` here as well, just so this reads more naturally. Currently, it's basically "the type of this page is the same as if it were in the posts category", but would be more logical to say "the type of this page is a post".

PR #361, opened 2023-11-01
Specify logo image sizes

alphapapa:
As discussed in our meeting today, this may slightly improve the Lighthouse scores.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -10,7 +10,7 @@
   <div class="container is-max-widescreen">
     <div id="footer-columns" class="columns">
       <div class="footer-logo-column">
-        <img id="footer-logo" src="{{ printf "/images/%s" $logo | relURL }}" alt="{{ $title }} logo. {{ $navbarLogo.altText }}">
+        <img id="footer-logo" src="{{ printf "/images/%s" $logo | relURL }}" alt="{{ $title }} logo. {{ $navbarLogo.altText }}" width="67" height="75">
       </div>
       {{- range $quickLinks }}
         <div class="link-column">
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -179,7 +179,8 @@ a > code {
 }
 
 #footer-logo {
-  height: 75px;
+  height: 4.75em;
+  width: auto;
 }
 
 .footer-column {
--- a/layouts/partials/hero.html
+++ b/layouts/partials/hero.html
@@ -12,7 +12,7 @@
           <div class="hero-title-content">
               <div class="hero-title">
                   {{ $title }}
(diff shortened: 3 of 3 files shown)


layouts/partials/footer.html
```diff
@@ -9,7 +9,7 @@
   <div class="container is-max-widescreen">
     <div id="footer-columns" class="columns">
       <div class="footer-logo-column">
-        <img id="footer-logo" src="{{ printf "/images/%s" $logo | relURL }}" alt="{{ $title }} logo. {{ $navbarLogo.altText }}">
+        <img id="footer-logo" src="{{ printf "/images/%s" $logo | relURL }}" alt="{{ $title }} logo. {{ $navbarLogo.altText }}" width="67" height="75">
```
stefanv, line 12: Probably best to use a scalable unit like `rem`.

alphapapa: Agreed, however HTML IMG tags specify sizes only in pixels.  The Lighthouse report recommended this change, which is the only reason for it.  But I'll see if it will accept specifying the size with a STYLE attribute in relative units.

alphapapa: Further research shows that, according to the resources linked by the Lighthouse report, the current best practice is to specify the image's expected width and height in the HTML tag in pixels, mainly for the sake of indicating the image's aspect ratio, and then using CSS properties to set the image's final size after loading.  (See, e.g. [this web.dev article](https://web.dev/articles/optimize-cls?utm_source=lighthouse&utm_medium=node#images_without_dimensions).

layouts/partials/footer.html
```diff
@@ -10,7 +10,7 @@
...
```
stefanv, line 13: Is the intent to still have explicit dimensions here, or is the CSS above supposed to take care of it?

stefanv:
An in-depth study of the issue, in case this comes up again: https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/

stefanv:
I think this solution is not generic, given that the logo is externally specified.

We could use https://gohugo.io/content-management/image-processing/#image-rendering, but it's a lot of work for fairly little benefit.

I'd say we either (a) punt on the issue or (b) require that logos are square and set the dimensions accordingly.

layouts/partials/footer.html
alphapapa: Yes, the explicit dimensions in the IMG tag are the recommended best practice.  It informs the browser of the image's aspect ratio early in the rendering process, which helps prevent re-rendering as resources are loaded.  Later, the CSS sets the image's actual display size.

jarrodmillman:
Punting on this for now, we can reopen if needed.

PR #362, opened 2023-11-01
Move shortcuts to top of container

stefanv:


jarrodmillman (approved):

stefanv:
Thanks for the top fix; forgot that sticky doesn't work without it.

PR #366, opened 2023-11-02
Use mathjax

jarrodmillman:


layouts/partials/head.html
```diff
@@ -0,0 +1,13 @@
+{{ if  or (eq .Section "posts") (eq .Section "tags") }}
```
stefanv, line 1: I'm not sure whether this is sufficient. Let's discuss.

jarrodmillman: Yes, we will need to update it. I was planning to bring everything over so the site would work and then clean up in a future PR.

jarrodmillman: Yes, we will need to update it. I was planning to bring everything over so the site would work and then clean up in a future PR.

stefanv (approved):

PR #387, opened 2023-11-06
Tabs tidy and shortcuts

stefanv:


stefanv:
Rebased on main.

stefanv:
I've noticed that our tabs do not behave all that well when the screen width is reduced dramatically. My last commit improves it a bit, but it is by no means perfect.

jarrodmillman (approved):

PR #388, opened 2023-11-06
Move font configuration to CSS; add documentation

### LINKED ISSUE #345: Remove font configuration option in config.yaml
Should be set via CSS.

Perhaps worth giving an example of how to do this in the docs.

alphapapa:
Closes #345.

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -373,10 +373,6 @@ a.headerlink {
   visibility: visible;
 }
 
-/* Local Variables: */
-/* css-indent-offset: 2 */
-/* End: */
-
 #breadcrumbs {
   margin: 0;
   padding: 0;
@@ -387,3 +383,7 @@ a.headerlink {
   font-style: italic;
   margin-top: -0.5rem;
 }
+
+/* Local Variables: */
+/* css-indent-offset: 2 */
+/* End: */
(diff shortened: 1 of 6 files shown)


stefanv:
@alphapapa Thanks for working through this example. Reading your PR, I realize I had a fundamental oversight, which is that if we remove the configuration option, the user loses the ability to load those fonts from Google, which means the CSS statement won't be helpful.

So, unless we came up with a way to also load the fonts, this plan was a dead-end from the get go. :disappointed: 

Sorry about that!

jarrodmillman:
I think we decided not to do this. If I am wrong, we can just reopen.

alphapapa:
> @alphapapa Thanks for working through this example. Reading your PR, I realize I had a fundamental oversight, which is that if we remove the configuration option, the user loses the ability to load those fonts from Google, which means the CSS statement won't be helpful.
> 
> So, unless we came up with a way to also load the fonts, this plan was a dead-end from the get go. 😞
> 
> Sorry about that!

No problem.

We could consider adding a bit of documentation showing how to add CSS that retrieves fonts from Google, and linking to the canonical documentation that shows how to do that.  Of course, it is more convenient to put the font name in the config file and have the CSS generated, so... :shrug:   :)

PR #391, opened 2023-11-06
Add: "More" links after post summaries

alphapapa:
After exploring a number of variants ("Read more" vs. "more", arrow vs. ellipsis, left-align vs. right-aligned, in a separate paragraph vs. immediately after the summary text, italic vs. normal, small-caps variants and non-, etc.), I settled on this.

I think it would look even better (less cluttered) were it not underlined, but which links to underline is probably a wider discussion that needs to be had (or one that has been had that I'm just not aware of).

Please let me know if you'd like me to make more changes.  Thanks.

Visible at: https://deploy-preview-391--scientific-python-hugo-theme.netlify.app/blog/

### DIFF
--- a/assets/theme-css/posts.css
+++ b/assets/theme-css/posts.css
@@ -86,8 +86,6 @@ div.MathJax_Display {
 }
 
 .read-more {
-  margin-top: 0.5em;
-  font-variant-caps: small-caps;
   font-style: italic;
 }
 
--- a/assets/theme-css/posts.css
+++ b/assets/theme-css/posts.css
@@ -85,6 +85,12 @@ div.MathJax_Display {
   margin: 0.5em;
 }
 
+.read-more {
+  margin-top: 0.5em;
+  font-variant-caps: small-caps;
+  font-style: italic;
+}
+
 .tag-cloud {
   position: sticky;
   align-self: flex-start;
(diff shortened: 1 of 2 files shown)


stefanv:
Examples of three variants:

image

stefanv:
For now, let's render the link inline and call it a day:

image

`display: inline` or `inline-block`.

jarrodmillman:
I think we should just follow the pydata-sphinx-theme example:
https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/blog/index.html

If we don't like it, I think we should make a PR to pydata-sphinx-theme and see if the community agrees. Right now, it is just the three of us working alone, but pydata-sphinx-theme has many developers from many projects providing feedback.

jarrodmillman:
If the two of you disagree, then I will go with whatever the two of you decide. If we all disagree, then I think we should go with what the community has already agreed to (i.e., the pydata-sphinx-theme styling).

stefanv (approved):

stefanv:
Thanks, Adam; I think your aesthetic input was good. It's a bit of a balancing game between doing what we think is optimal and getting ourselves in alignment with the PDST. I guess we're leaning a bit towards that side right now, but fortunately we have room to adjust in the future.

I see the everything clickable underlined trend is now on GitHub as well, so I guess we'll ride it out and see where it lands. Our hope is to ultimately have styling decisions being made on PDST, and automatically propagate here.  I.e., not having to worry about it :)

PR #393, opened 2023-11-06
Use pandoc markdown renderer

jarrodmillman:


jarrodmillman:
Seems like this isn't well supported.

PR #402, opened 2023-11-09
Fix: (posts, shortcuts) Adjust width, padding for narrower screens

alphapapa:
This prevents the main content container from being directly against
the edge of the screen or being pushed off-screen, and prevents the
shortcuts container from being too narrow.

<details><summary>Screenshots</summary>
<p>

before
after

</p>
</details>

### DIFF
--- a/assets/theme-css/content.css
+++ b/assets/theme-css/content.css
@@ -1,38 +1,21 @@
-/* Default: single column layout for mobile */
-.content-container {
-  max-width: 70em;
-  margin: 0 30px;
-  flex-direction: column;
-}
-
-/* Full screen two-column layout for desktop etc. */
-@media only screen and (min-width: 75em) {
-  .content-container {
-    flex-basis: 70em;
-    flex-direction: row;
-  }
-
-  article {
-    flex-grow: 1;
-    flex-shrink: 0;
-    flex-basis: calc(50% - 1rem);
-  }
-}
-
-/* Other rules. */
-
 .content-padding {
   display: flex;
+  flex-wrap: nowrap;
   justify-content: center;
   margin: 1em;
 }
 
-.content-container .divider {
-  height: 3px;
-  border-radius: 50px;
-  background: var(--pst-color-text-base);
-  width: 60px;
-  &.is-centered {
-    margin: 0 auto;
+.content-container {
+  max-width: 70em;
+  flex-basis: 70em;
+  flex-shrink: 1;
+  margin: 0 30px;
+  flex-direction: column;
+}
+
+/* Start wrapping at the same breakpoint as the one that enables the burger menu */
+@media screen and (max-width: 1023px) {
+  .content-padding {
+    flex-wrap: wrap;
   }
 }
--- a/assets/theme-css/content.css
+++ b/assets/theme-css/content.css
@@ -8,7 +8,7 @@
 /* Full screen two-column layout for desktop etc. */
 @media only screen and (min-width: 75em) {
   .content-container {
-    min-width: 70rem;
+    flex-basis: 70em;
     flex-direction: row;
   }
 
--- a/assets/theme-css/content.css
+++ b/assets/theme-css/content.css
@@ -6,16 +6,11 @@
 }
 
 .content-container {
-  max-width: 70em;
   flex-basis: 70em;
   flex-shrink: 1;
-  margin: 0 30px;
   flex-direction: column;
-}
+  margin: 0 30px;
 
-/* Start wrapping at the same breakpoint as the one that enables the burger menu */
-@media screen and (max-width: 1023px) {
-  .content-padding {
-    flex-wrap: wrap;
-  }
+  /* Handle code cells overflowing */
+  overflow: auto;
 }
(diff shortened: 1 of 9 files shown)


stefanv:
Removing `min-width` is not quite right here, as it introduces problems with narrow pages (see, e.g, the examples / kitchen sink page).

What is happening is that we are not being careful about the specifications. We ask for 70em minimum body width, so the only element that can shrink is the shortcut. If you shrink more, you hit the media selector, which enables flex box layouts and then everything is fine.

So, what we should probably do is either (a) make the media selector act a bit more quickly or (b) always lay out the page using flex, making use of flex-basis and flex-shrink (the latter set to 0 for the shortcuts container).

jarrodmillman:
@alphapapa Let's discuss this in a call. I have some ideas / questions about how to get the page layouts right.

stefanv:
I think what you want is flex-basis: 70em;

.dir-locals.el
```diff
@@ -2,4 +2,4 @@
 ;;; For more information see (info "(emacs) Directory Variables")
 
 ((css-mode . ((css-indent-offset . 2)))
- (html-mode . ((sgml-basic-offset . 4))))
```
jarrodmillman, line 5: Would you mind adding this to a separate PR. I don't mind adding code for emacs, but let's leave it in separate PRs.

alphapapa:
@stefanv You were right about using flex-basis and such.  The problem was that one of the `pre` elements on the Features page had a very long line, and even when its `overflow` was set to scroll, and even when I constrained its max width, the browser was using its theoretical ideal width to calculate the minimum width of the parent flex element, which was pushing the shortcuts container off the screen.  Once I set the `pre` element's wrapping and breaking properties, suddenly it was rendered sensibly again, and then all that remained was to do as you suggested.  Thanks!

alphapapa:
This diff seems to also help with the content jumping to the middle of the page as the viewport narrows to a certain point, e.g. on the Examples page:
```diff
diff --git a/assets/theme-css/bulma.css b/assets/theme-css/bulma.css
index 5a91d86..cf15d2f 100644
--- a/assets/theme-css/bulma.css
+++ b/assets/theme-css/bulma.css
@@ -1020,11 +1020,3 @@ a.navbar-item:hover {
 .hero ul {
   border-bottom: none;
 }
-.section {
-  padding: 3rem 1.5rem;
-}
-@media screen and (min-width: 1024px) {
-  .section {
-    padding: 3rem 3rem;
-  }
-}
```
But even that isn't enough, as at a certain point, it jumps back to the center again because of this rule:
```diff
diff --git a/assets/theme-css/content.css b/assets/theme-css/content.css
index 3e54ab8..82c12ca 100644
--- a/assets/theme-css/content.css
+++ b/assets/theme-css/content.css
@@ -23,7 +23,6 @@
 
 .content-padding {
   display: flex;
-  justify-content: center;
   margin: 1em;
 }
 ```
Those also have some other minor side effects with regard to margins and spacing.

Anyway, more work remains to be done with regard to shrinking viewports affecting page layout in surprising ways.

File: assets/theme-css/content.css [hunk 2]
```diff
@@ -1,15 +1,32 @@
...
+  .content-container {
+    flex-basis: 70em;
+    flex-direction: row;
+  }
+
+  article {
+    flex-grow: 1;
+    flex-shrink: 0;
+    flex-basis: calc(50% - 1rem);
+  }
```
stefanv, line 19: I wish we had a comment to explain why we needed this CSS :thinking:

assets/theme-css/pst/base/_base.scss
```diff
@@ -160,6 +160,11 @@ pre {
   border: 1px solid var(--pst-color-border);
   border-radius: $admonition-border-radius;
 
+  // Prevent long, monospaced lines and words in PRE elements from
+  // breaking page layout.  Very non-obvious problem and solution.
+  white-space: pre-wrap;
+  word-break: break-all;
+
```
stefanv, line 167: Could you add a comment to explain why this is the correct solution? That would be helpful to reviewers now, and to our future selves.

assets/theme-css/content.css
```diff
@@ -1,15 +1,32 @@
+/* Default: single column layout for mobile */
+.content-container {
+  max-width: 70em;
+  margin: 0 30px;
+  flex-direction: column;
+}
+
+/* Full screen two-column layout for desktop etc. */
+@media only screen and (min-width: 75em) {
```
stefanv, line 9: If the flex basis for this one is 70em, and for the shortcuts 15em, then I suppose we want to activate this at 85em?

assets/theme-css/shortcuts.css
```diff
@@ -2,8 +2,11 @@
   position: sticky;
   align-self: flex-start;
   top: 5rem;
-  width: 100%;
-  max-width: 150px;
+  flex-basis: 15em;
+  /* We disable shrinking for this container, otherwise the single-line
+   items start wrapping to two lines, especially when active and
+   bolded, which causes distracting layout changes. */
+  flex-shrink: 0;
```
stefanv, line 9: :+1:

stefanv:
The blog looks better now; not sure if it will be 100% on the preview render, since we also have to merge https://github.com/scientific-python/blog.scientific-python.org/pull/167

jarrodmillman (approved):

jarrodmillman (approved):

Issue #397, opened 2023-11-07
Missing documention


jarrodmillman:
- [ ] dart-sass

alphapapa:
I've made https://github.com/scientific-python/scientific-python-hugo-theme/pull/401 which updates the installation instructions.  As it is now, it just refers users to Hugo's own installation instructions, which seem adequate; I doubt that we'd want to reiterate theirs (and have to keep ours updated if theirs change).  If you'd like me to add more detail or make other changes, please let me know.

PR #403, opened 2023-11-09
Docs: Add tabs shortcode example

alphapapa:
See <https://github.com/scientific-python/scientific-python-hugo-theme/issues/397>.

This also adds a `shortcode-example` shortcode, only used in our documentation, which distinguishes the rendered shortcode examples from the content in the page describing them.

It also fixes a small bug (encountered when adding the tabs shortcode example) regarding nested Markdown code blocks in the generated documentation.

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -270,10 +270,6 @@ a.headerlink {
   visibility: visible;
 }
 
-/* Local Variables: */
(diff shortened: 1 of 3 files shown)


jarrodmillman:
Would put the style change stuff in a separate PR? Let's keep this one focused on just adding the tabs shortcode example.

jarrodmillman (approved):

PR #411, opened 2023-11-14
Remove bulma box-sizing

jarrodmillman:
I've been thinking about removing this for awhile. I hadn't been able to find where it changes anything, but now removing it fixes an issue with numpy.org.

### DIFF
--- a/assets/theme-css/bulma.css
+++ b/assets/theme-css/bulma.css
@@ -138,9 +138,6 @@ input,
 select {
   margin: 0;
 }
-html {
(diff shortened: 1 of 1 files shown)


assets/theme-css/bulma.css
```diff
@@ -141,11 +141,6 @@ select {
 html {
```
jarrodmillman, line 141: I am going to try to remove this as well. Just testing one thing at a time.

stefanv (approved):

PR #418, opened 2023-11-16
Add copy button for code blocks

alphapapa:
This implements a simple, bespoke button attached to code blocks that copies the code to the clipboard.  (Rather than use, e.g. [clipboard.js](https://clipboardjs.com/), which is 3 KB gzipped.)

It looks like this:

<details><summary>Details</summary>
<p>

Screen Shot 2023-11-16 at 17 17 29

</p>
</details> 

It still needs a bit of visual tweaking in the CSS for, e.g. one-line, very wide code blocks, but that shouldn't take much more work.

### DIFF
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -20,9 +20,8 @@ function add_button_to(element) {
 }
 
 function copy_content_of(event) {
-    // Omit "Copy" prefix from the button itself.
-    target = this.parentElement.textContent.substring(4);
-    navigator.clipboard.writeText(target)
+    content = this.parentElement.parentElement.textContent;
+    navigator.clipboard.writeText(content)
         .then(() => update_button(this));
 }
 
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -11,26 +11,25 @@ const pre_selector = "div.highlight pre";
 document.querySelectorAll(pre_selector).forEach(add_button_to);
 
 function add_button_to(element) {
-    div = document.createElement("div");
-    button = document.createElement("button");
-    div.classList.add("copy-button");
-    button.addEventListener("click", copy_content_of);
-    div.append(button);
-    element.prepend(div);
+  div = document.createElement("div");
+  button = document.createElement("button");
+  div.classList.add("copy-button");
+  button.addEventListener("click", copy_content_of);
+  div.append(button);
+  element.prepend(div);
 }
 
 function copy_content_of(event) {
-    content = this.parentElement.parentElement.textContent;
-    navigator.clipboard.writeText(content)
-        .then(() => update_button(this));
+  content = this.parentElement.parentElement.textContent;
+  navigator.clipboard.writeText(content).then(() => update_button(this));
 }
 
-function update_button(button, new_text="Copied  ", reset=true) {
-    old_text = button.innerText;
-    button.innerText = new_text;
-    if (reset) {
-        setTimeout(update_button, 2000, button, old_text, false);
-    }
+function update_button(button, new_text = "Copied  ", reset = true) {
+  old_text = button.innerText;
+  button.innerText = new_text;
+  if (reset) {
+    setTimeout(update_button, 2000, button, old_text, false);
+  }
 }
 
 // copy-code-blocks.js ends.
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -11,11 +11,12 @@ const pre_selector = "div.highlight pre";
 document.querySelectorAll(pre_selector).forEach(add_button_to);
 
 function add_button_to(element) {
+    div = document.createElement("div");
+    div.classList.add("copy-button");
     button = document.createElement("button");
-    button.classList.add("copy-button")
-    button.append(document.createTextNode("Copy"));
     button.addEventListener("click", copy_content_of);
-    element.prepend(button);
+    div.append(button);
+    element.prepend(div);
 }
 
 function copy_content_of(event) {
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -24,11 +24,9 @@ function copy_content_of(event) {
   navigator.clipboard.writeText(content).then(() => update_button(this));
 }
 
-function update_button(button, new_text = "Copied  ", reset = true) {
-  button.innerText = new_text;
-  if (reset) {
-    setTimeout(update_button, 2000, button, "", false);
-  }
+function update_button(button, clicked_class = "clicked", timeout_ms = 2000) {
+  button.classList.add(clicked_class);
+  setTimeout(() => button.classList.remove(clicked_class), timeout_ms);
 }
 
 // copy-code-blocks.js ends.
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -1,0 +1,40 @@
+// copy-code-blocks.js
+// Copyright: 2023  Scientific Python Developers
+// Author: Adam Porter <adam@alphapapa.net>
+// License: BSD 3-Clause License
+//   see <https://raw.githubusercontent.com/scientific-python/scientific-python-hugo-theme/main/LICENSE>
+
+// This file implements a button attached to code blocks which copies
+// the code to the clipboard.
+
+// This selector should match the PRE elements, each of which may
+// itself contain the text to be copied, or may contain a CODE element
+// which contains the text to be copied.
+const pre_selector = "div.highlight pre";
+
+document.querySelectorAll(pre_selector).forEach(add_button_to);
+
+function add_button_to(element) {
+    button = document.createElement("button");
+    button.classList.add("copy-button")
+    button.append(document.createTextNode("Copy"));
+    button.addEventListener("click", copy_content_of);
+    element.prepend(button);
+}
+
+function copy_content_of(event) {
+    // Omit "Copy" prefix from the button itself.
+    target = this.parentElement.textContent.substring(4);
+    navigator.clipboard.writeText(target)
+        .then(() => update_button(this) );
+}
+
+function update_button(button, new_text="Copied", reset=true) {
+    old_text = button.innerText;
+    button.innerText = new_text;
+    if (reset) {
+        setTimeout(update_button, 1000, button, old_text, false);
+    }
+}
+
+// copy-code-blocks.js ends.
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -12,8 +12,8 @@ document.querySelectorAll(pre_selector).forEach(add_button_to);
 
 function add_button_to(element) {
     div = document.createElement("div");
-    div.classList.add("copy-button");
     button = document.createElement("button");
+    div.classList.add("copy-button");
     button.addEventListener("click", copy_content_of);
     div.append(button);
     element.prepend(div);
@@ -23,7 +23,7 @@ function copy_content_of(event) {
     // Omit "Copy" prefix from the button itself.
     target = this.parentElement.textContent.substring(4);
     navigator.clipboard.writeText(target)
-        .then(() => update_button(this) );
+        .then(() => update_button(this));
 }
 
 function update_button(button, new_text="Copied  ", reset=true) {
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -20,7 +20,7 @@ function add_button_to(element) {
 }
 
 function copy_content_of(event) {
-  content = this.parentElement.parentElement.textContent;
+  content = this.parentElement.parentElement.textContent + "\n";
   navigator.clipboard.writeText(content).then(() => update_button(this));
 }
 
--- a/assets/js/copy-code-blocks.js
+++ b/assets/js/copy-code-blocks.js
@@ -26,11 +26,11 @@ function copy_content_of(event) {
         .then(() => update_button(this) );
 }
 
-function update_button(button, new_text="Copied", reset=true) {
+function update_button(button, new_text="Copied  ", reset=true) {
     old_text = button.innerText;
     button.innerText = new_text;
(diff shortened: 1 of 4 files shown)


assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
+// copy-code-blocks.js
+// Copyright: 2023  Scientific Python Developers
+// Author: Adam Porter <adam@alphapapa.net>
+// License: BSD 3-Clause License
+//   see <https://raw.githubusercontent.com/scientific-python/scientific-python-hugo-theme/main/LICENSE>
+
```
stefanv, line 6:
```suggestion
```

It's a standard practice on our projects not to include individual author affiliations.
The license file then applies to the entire project.

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
...
+// This file implements a button attached to code blocks which copies
+// the code to the clipboard.
+
+// This selector should match the PRE elements, each of which may
+// itself contain the text to be copied, or may contain a CODE element
+// which contains the text to be copied.
+const pre_selector = "div.highlight pre";
```
stefanv, line 13: Double check that this also catches fenced code blocks that don't specify a language.

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
...
+
+document.querySelectorAll(pre_selector).forEach(add_button_to);
+
+function add_button_to(element) {
+    button = document.createElement("button");
+    button.classList.add("copy-button")
+    button.append(document.createTextNode("Copy"));
+    button.addEventListener("click", copy_content_of);
```
stefanv, line 21: This should be modeled after PDST, i.e. use a FontAwesome icon, and then when clicked display the Copied! text.

https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/blocks.html

Fading effects are not required.

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
...
+    element.prepend(button);
+}
+
+function copy_content_of(event) {
+    // Omit "Copy" prefix from the button itself.
+    target = this.parentElement.textContent.substring(4);
+    navigator.clipboard.writeText(target)
+        .then(() => update_button(this) );
```
stefanv, line 29:
```suggestion
        .then(() => update_button(this));
```

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
...
+    button = document.createElement("button");
+    button.classList.add("copy-button")
+    button.append(document.createTextNode("Copy"));
+    button.addEventListener("click", copy_content_of);
+    element.prepend(button);
+}
+
+function copy_content_of(event) {
+    // Omit "Copy" prefix from the button itself.
+    target = this.parentElement.textContent.substring(4);
```
stefanv, line 27: What is the substring(4) used for?

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,40 @@
...
+    navigator.clipboard.writeText(target)
+        .then(() => update_button(this) );
+}
+
+function update_button(button, new_text="Copied", reset=true) {
```
stefanv, line 32: no need for a reset argument here.

stefanv (commented):
Thanks, this is already reasonably close!

assets/js/copy-code-blocks.js
alphapapa: Understood.  Thanks.

Because of the way Hugo concats local JS files together, I left the filename comment on the first and last lines, to make its contents easier to distinguish in the bundled source code.  But I can remove that too if you like.

assets/js/copy-code-blocks.js
alphapapa: That does look nice.

I implemented something a bit simpler to start with, please see https://github.com/scientific-python/scientific-python-hugo-theme/pull/418/commits/26ad2f903664d285b9db33702d82890e1708a9dd and https://github.com/scientific-python/scientific-python-hugo-theme/pull/418/commits/401e4c7141e0e50e3ea64f56b8fa960f110723fe and let me know if you'd like more changes to it.

assets/js/copy-code-blocks.js
alphapapa: It was sort of an elegant hack, but no longer needed after other changes.

assets/js/copy-code-blocks.js
alphapapa: I'm not sure what you mean.

assets/js/copy-code-blocks.js
alphapapa: Yes, testing on all the code blocks on the `shortcodes/` page, ones both with and without a language are copied correctly.  Thanks.

alphapapa:
pre-commit.ci autofix

stefanv:
Let's get rid of the boundary of the button, add some text when overed over (Copy), and use the empty copy icon (https://fontawesome.com/icons/copy?f=classic&s=regular). Then probably good enough to merge.

Please check under light and dark mode; dark mode is currently jarring because of button background.

stefanv:
One more thing: we should copy a newline after the end of the content, since this is all multi-line content blocks.

assets/js/copy-code-blocks.js
stefanv: You never use `update_button` with anything other than `reset=true`.

assets/theme-css/pst/content/_code.scss
```diff
@@ -12,6 +12,24 @@ div.literal-block-wrapper {
   flex-direction: column;
   width: unset;
   border-radius: $admonition-border-radius;
+
+  // Copy-to-clipboard button.
+  & .copy-button {
```
stefanv, line 17: Note: we cannot modify the pst theme, so this will have to go elsewhere. Better if we can re-use what they have, whenever possible.

alphapapa: Ah, sorry, I keep accidentally editing the wrong files this way.   :)  Will fix.  Thanks.

assets/js/copy-code-blocks.js
alphapapa: The function calls itself again to reset the button from the `setTimeout` call.

assets/js/copy-code-blocks.js
```diff
@@ -0,0 +1,35 @@
...
+function copy_content_of(event) {
+  content = this.parentElement.parentElement.textContent;
+  navigator.clipboard.writeText(content).then(() => update_button(this));
+}
+
+function update_button(button, new_text = "Copied  ", reset = true) {
+  old_text = button.innerText;
+  button.innerText = new_text;
+  if (reset) {
+    setTimeout(update_button, 2000, button, old_text, false);
```
stefanv, line 31:
```suggestion
    setTimeout(() => update_button(button, old_text, false), 2000);
```

Check that this still works if clicking the button twice in rapid succession.

layouts/partials/css.html
```diff
@@ -21,34 +21,33 @@
 <!-- Fallback font for symbols (such as "🛈"). --
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Noto+Sans+Symbols+2">
 
-<!-- SASS -->
-{{- $sass := (slice "theme-css/pst/pydata-sphinx-theme.scss") | append (resources.Match "css/*.scss") -}}
+<!-- Process and include Sass files. -->
+{{- $sass := (slice "theme-css/pst/pydata-sphinx-theme.scss")
+           | append (resources.Match "theme-css/*.scss")
```
stefanv, line 26: `theme-css/**/*.scss`?

stefanv: or `*-css/**/*.scss`, then you avoid the empty slices too.

assets/js/copy-code-blocks.js
stefanv: What did you think of this syntax?

alphapapa: Good catch.  Fixed.  Thanks.

stefanv: And what did you think of the bracket function syntax? I think it reads a bit more easily.

layouts/partials/css.html
alphapapa: I'm not sure, but I've spun out that commit to https://github.com/scientific-python/scientific-python-hugo-theme/pull/429.

alphapapa:
pre-commit.ci autofix

alphapapa:
Note as well that this PR is based on https://github.com/scientific-python/scientific-python-hugo-theme/pull/429, which should be merged first.

assets/theme-css/code.scss
```diff
@@ -0,0 +1,44 @@
...
+    visibility: hidden;
+    z-index: 1;
+  }
+  &:hover .copy-button {
+    visibility: visible;
+  }
+  & .chroma .copy-button button {
+    // Fix alignment that the .chroma rules affect.
+    padding: 0.5rem;
+  }
```
stefanv, line 23: The PR looks good overall; is there a reason we cannot adopt the styling from PDST directly? For bootstrap styling, we have to add it, but theirs should already be available to us.

layouts/partials/css.html
stefanv, line 26: Presumably these changes now happen in another PR.

assets/theme-css/code.scss
alphapapa: I spent about 30 minutes trying to make that work.  The source of the CSS used is obfuscated by being in a Sphinx extension.  So I copied the generated CSS and tried to make that work by adjusting the HTML and CSS and JS, and the differences seemed too great to be worth reconciling; to get the same appearance was going to take a lot more fiddling.  But if you want me to spend more time to try to make that work, just let me know.

layouts/partials/css.html
alphapapa: Yes, in https://github.com/scientific-python/scientific-python-hugo-theme/pull/429

assets/theme-css/code.scss
stefanv: That's fine, I was more wondering whether we could keep it, without adjustment, and simply render the same elements as they do. If that's not feasible, we'll just put this in as-is.

layouts/partials/css.html
stefanv: Merged, so this can be rebased.

assets/theme-css/code.scss
alphapapa: I'll quickly take another look and see if there's a simple approach that I overlooked to make it work that way; if not, I'll merge this as-is.  Thanks.

alphapapa: Ok, I've pushed another branch that brings in the relevant Sass and CSS files from PST directly and then overrides the necessary properties in a separate file.  This link shows the difference between this pull request's branch and the branch that uses "more PST": https://github.com/alphapapa/scientific-python-hugo-theme/compare/wip/code-copy-button...alphapapa:scientific-python-hugo-theme:wip/code-copy-button-more-pst?expand=1  The rendering in the browser is the same; the only difference is the amount of code, and a small tweak to the JavaScript.

Please let me know which branch you prefer.  Thanks.

stefanv: Thanks @alphapapa. I like that approach very much.

Note that you don't need to manually copy over the PDST css, you can uncomment the import in `pydata-sphinx-theme.scss`.

alphapapa: > Thanks @alphapapa. I like that approach very much.

Ok, I'll see if I can tidy it up a bit more before we merge it.

> Note that you don't need to manually copy over the PDST css, you can uncomment the import in `pydata-sphinx-theme.scss`.

AFAICT uncommenting the `@import "./extensions/copybutton";` line didn't do anything because there was no such file to be imported, because the `_copybutton.scss` file had not been copied from the PDST repo.  As well, the PDST has two relevant files: the `_copybutton.scss` file from the `sphinx-copybutton` extension, and its own customizations in the `copybutton.scss` file.  After copying the first file, the `@import` line worked.

The disadvantage of this approach is that we end up copying more CSS files and rules than we need, and we end up overriding many of the CSS properties to fit our HTML; so if the PDST changes these rules in the future, and we import the files again, we'll likely have to update our overrides again to compensate.

But I'll see if I can harmonize our HTML a bit more with PDST to reduce the number of overrides needed.

jarrodmillman:
@alphapapa and I reviewed using more of the PST styling and decided it was too much code and hassle. So we are just going to stick with the current approach.

jarrodmillman:
This should be ready to go once it has been rebased and the conflict has been resolved.

assets/theme-css/code.scss
stefanv: It's worth keeping the overarching goal in mind here: to get rid of all our own styling.
So that, ultimately, we will have *all* of PDST SCSS in our repo, and almost none of our own.
It feels a bit clumsy now, but it fits better with the future goal.

Your goal of harmonizing our HTML a bit more with theirs is spot-on; thanks!

alphapapa: > It's worth keeping the overarching goal in mind here: to get rid of all our own styling. So that, ultimately, we will have _all_ of PDST SCSS in our repo, and almost none of our own. It feels a bit clumsy now, but it fits better with the future goal.
> 
> Your goal of harmonizing our HTML a bit more with theirs is spot-on; thanks!

Agreed, of course.  :)  After Jarrod and I compared the CSS in these two branches, we decided to go with the first branch in this PR so we can go ahead and get this merged.  

But I'll make another to-do item to see if I can find a way to finish harmonizing the HTML and remove all local CSS overrides that sit on top of the PDST CSS so its code can be used exclusively and as-is.  It seems that it will require going over with a fine-toothed comb to find out what's causing the button to not display the same way on SPHT as it does on PDST even with (what seems to be) the same HTML and CSS.

alphapapa:
> This should be ready to go once it has been rebased and the conflict has been resolved.

@jarrodmillman Rebased and pushed.  Thanks.

PR #429, opened 2023-11-29
Change/Fix: Include "theme-css/*.scss" files

alphapapa:
This includes Sass files in the `theme-css` directory (previously they were only included from the downstream `css` directory).

To be used in <https://github.com/scientific-python/scientific-python-hugo-theme/pull/418>.

### DIFF
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -23,8 +23,8 @@
 
 <!-- SASS -->
 {{- $sass := (slice "theme-css/pst/pydata-sphinx-theme.scss")
-           | append (resources.Match "theme-css/*.scss")
-           | append (resources.Match "css/*.scss") -}}
+           | append (resources.Match "theme-css/**/*.scss")
+           | append (resources.Match "css/**/*.scss") -}}
 {{- range $sass -}}
 
 {{ with . }}  <!-- Skips nil elements from appending empty resources.Match slices. -->
@@ -41,8 +41,8 @@
 {{- end -}}
 {{- end -}}
 
-{{- $themeCssFiles := resources.Match "theme-css/*.css" -}}
-{{- $userCssFiles := resources.Match "css/*.css" -}}
+{{- $themeCssFiles := resources.Match "theme-css/**/*.css" -}}
+{{- $userCssFiles := resources.Match "css/**/*.css" -}}
 {{- $cssFiles := $themeCssFiles | append $userCssFiles }}
 
 {{- range $cssFiles -}}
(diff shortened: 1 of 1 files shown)


stefanv:
Note the review comment on `**/*` syntax being supported.

alphapapa:
> Note the review comment on `**/*` syntax being supported.

@stefanv Thanks, I applied that glob pattern to all the relevant places in this file.

alphapapa:
@stefanv Using the `**/*` glob pattern caused `Undefined mixin` errors which broke the build.  I'm not sure what's going on there, but after I reverted those changes in https://github.com/scientific-python/scientific-python-hugo-theme/pull/429/commits/72382e92fe617b5f984a3153517c0cfa9214fcb6, the build succeeded again.

layouts/partials/css.html
```diff
@@ -21,34 +21,35 @@
...
-<link rel="stylesheet" href="{{ $custom_style.RelPermalink }}" integrity="{{ $custom_style.Data.Integrity }}">
-{{- end -}}
-
+  {{ if $inServerMode -}}
+    {{ $custom_style := . | resources.ExecuteAsTemplate . $page -}}
+    <link rel="stylesheet" href="{{ $custom_style.RelPermalink }}">
+  {{ else }}
+    {{ $custom_style := . | resources.ExecuteAsTemplate . $page | minify | fingerprint -}}
+    <link rel="stylesheet" href="{{ $custom_style.RelPermalink }}" integrity="{{ $custom_style.Data.Integrity }}">
+  {{- end -}}
```
stefanv, line 54: The reason for this indentation is because of the generated HTML.
I'm OK with keeping it as-is, but my emacs also doesn't indent that way by default.
Perhaps you have settings I can adopt?

stefanv:
> @stefanv Using the `**/*` glob pattern caused `Undefined mixin` errors which broke the build. I'm not sure what's going on there, but after I reverted those changes in [72382e9](https://github.com/scientific-python/scientific-python-hugo-theme/commit/72382e92fe617b5f984a3153517c0cfa9214fcb6), the build succeeded again.
> 
> #418 is waiting on this to be merged, so if you'd like me to look into the issues with that glob pattern, I could do that in a separate issue.

It probably means that we should only load top-level PDST SCSS, which then itself imports deeper down.

layouts/partials/css.html
alphapapa: I indented these lines manually for clarity in the source code (the `{{- end -}}` repeated 3 lines in a row, with differing opening lines for each block, is hard to follow).  But just let me know how you want this to be done.

stefanv (approved):

PR #444, opened 2023-12-19
Refactor bulma navbar css

jarrodmillman:


stefanv:
I think now that we have only one navbar, we may be able to get rid of the transparent classes?

jarrodmillman:
I am going to do this in several stages to make it easier to follow. I will also try to get rid of is-fresh. And then move everything into navbar.css.

stefanv:
Thanks, Jarrod! I noticed an (existing) artifact, which is that when you switch between light and dark mode, the navbar fades in. Yet another animation to get rid of!

stefanv:
The hamburger bar menu highlight also changed (or, at least, is different from what I see now on scipy.org). I'll let you decide whether that's important to keep or not; go ahead an merge otherwise.

PR #458, opened 2024-01-26
Simplify footer

jarrodmillman:


stefanv:
This changes the footer color. I quite like the current one blue version?

PR #459, opened 2024-01-26
Increase color contrast in footer

### LINKED ISSUE #457: Color contrast on page footer
Hi folks, when checking the page recently I noticed that the hover/highlight color on links gives a low contrast on the page footer:

Screenshot_20240125_213738

As a quick check, I got the following on https://coolors.co/contrast-checker/8045e5-013243

Screenshot_20240125_213308

I believe the violet is inherited from the PyData Sphinx Theme, so not sure what the right solution would be here.

jarrodmillman:
Fix #457. I am using the same hoover color as used in the `about` button and news banner.

2024-01-26T08:22:48,295534305-08:00

https://coolors.co/contrast-checker/4dabcf-013243

### DIFF
--- a/assets/theme-css/footer.css
+++ b/assets/theme-css/footer.css
@@ -35,7 +35,7 @@
 }
 
 .footer-item a:hover {
-  color: var(--pst-color-link-hover);
+  color: var(--colorPrimaryLight);
 }
 
 .footer-actions {


stefanv (approved):

jarrodmillman:
I am going to merge this and close the issue. This is much better than what we had, but we still need to revisit this when we do our accessibility audit.

melissawm:
Thanks for the swift response!

PR #461, opened 2024-01-26
Fix navbar / active burger issue (#460)

### LINKED ISSUE #460: Navbar / active burger bug
If you make the screen small, click on the burger, and make the screen big again, you get this
2024-01-26T09:41:59,478017999-08:00

It should look like this
2024-01-26T09:43:25,796774389-08:00

jarrodmillman:
Fix #460.

### DIFF
--- a/assets/theme-css/navbar.css
+++ b/assets/theme-css/navbar.css
@@ -116,18 +116,6 @@ nav,
   color: var(--colorPrimaryLight);
 }
 
-.navbar-menu.is-active {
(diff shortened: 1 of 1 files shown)


stefanv:
Do you know why moving this into the media selector makes a difference?

jarrodmillman:
> Do you know why moving this into the media selector makes a difference?

My thinking was that it should only be defined on small screens (when we are using the burger).

PR #463, opened 2024-01-31
Add PST cards

jarrodmillman:


layouts/shortcodes/card.html
```diff
@@ -0,0 +1,37 @@
...
+
+*/}}
+
+<div class="bd-content">
+<div class="sd-card sd-shadow-sm">
+  {{ if .Get "title" }}
+  <div class="sd-card-title sd-font-weight-bold">
+    {{ .Get "title" }}
+  </div>
+  {{ end }}
```
stefanv, line 30:
```suggestion
  {{ with .Get "title" }}
  <div class="sd-card-title sd-font-weight-bold">
    {{ . }}
  </div>
  {{ end }}
```

layouts/shortcodes/card.html
```diff
@@ -0,0 +1,53 @@
...
+{{< /card >}}
+
+
+*/}}
+
+<div class="bd-content">
+<div class="sd-card sd-shadow-sm sd-mb-3">
+{{ with .Get "header" }}
+<div class="sd-card-header">
+  {{ . }}
```
stefanv, line 34: Should the header be run through markdownify? Same for footer.

stefanv (approved):
LGTM; only the one question.

PR #466, opened 2024-02-01
Add PST cards in grid

jarrodmillman:
Depends on #463 .

Maybe the cards information should be in a yaml file or something?
Then the grid shortcode could take the yaml file as the only input.

The cards may have a lot of features (buttons, headers, footers, titles, etc...)

pydata-sphinx-theme seems to just talk about cards in grids.
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#cards
https://github.com/pydata/pydata-sphinx-theme/blob/main/docs/index.md
https://github.com/pydata/pydata-sphinx-theme/blob/main/docs/user_guide/web-components.rst

sphinx-design has both.
https://sphinx-design.readthedocs.io/en/latest/cards.html
https://sphinx-design.readthedocs.io/en/latest/grids.html

I am thinking we could use the grid of cards to replace:
https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/layouts/partials/panel.html
https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/layouts/partials/keyfeatures.html

### DIFF
--- a/assets/theme-css/pst-deps/sphinx-design/_grid.scss
+++ b/assets/theme-css/pst-deps/sphinx-design/_grid.scss
@@ -1,0 +1,21 @@
+.sd-container,
+.sd-container-fluid,
+.sd-container-lg,
+.sd-container-md,
+.sd-container-sm,
+.sd-container-xl {
+  margin-left: auto;
+  margin-right: auto;
+  padding-left: var(--sd-gutter-x, 0.75rem);
+  padding-right: var(--sd-gutter-x, 0.75rem);
+  width: 100%;
+}
+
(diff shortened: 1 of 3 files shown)


stefanv:
I like the yaml idea, FWIW, otherwise specifying the cards soon becomes unwieldy.

And instead of storing that information in a file, it can appear directly in the shortcode body. See https://gohugo.io/functions/transform/unmarshal/

jarrodmillman:
This isn't ready, but I am merging so @brianhawthorne can refactor and fix it.

PR #469, opened 2024-02-08
Implement clickable cards

jarrodmillman:


stefanv:
The behavior I see right now is, when I hover over the cards, the two side-borders disappear, and the text wobbles a bit.

PR #474, opened 2024-02-13
Add FA icons for dropdowns

stefanv:
CSS/classes need to be cleaned up.

Question: do we want "animated" locks, i.e. one for open and one for closed? Would be quite easy to add with some additional CSS.


jarrodmillman:
We can hold off on the animated locks for now and add them in a second PR. I will make a follow-up issue.

jarrodmillman (approved):

Issue #480, opened 2024-02-15
Implement PST "buttons"


jarrodmillman:
For now we will just implement PST buttons using links (they do the same thing):
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#badges-and-button-links

The motivating case is to replace our panel partial with a grid of cards with buttons:
https://learn.scientific-python.org/

It may be worth taking a look at the underlying sphinx-design buttons (since we will probably want to use the "outline" version of the buttons like we currently do in the panel partial):
https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#buttons

jarrodmillman:
We can investigate using actual buttons and the correct [ARIA attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role). For now, we just need links that look like buttons.

If we end up needing actual buttons for forms and things, we could try to do something similar to what bootstrap does:
https://getbootstrap.com/docs/4.0/components/buttons/

PR #483, opened 2024-02-15
Add button shortcode

alphapapa:
Implements for https://github.com/scientific-python/scientific-python-hugo-theme/issues/480


jarrodmillman:
Could you add an example of and outline version? There aren't examples on the PST page, but I think they should work just like the outline examples here:
https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html

jarrodmillman (approved):

PR #498, opened 2024-02-22
Remove panel

jarrodmillman:
Use a grid of cards instead.


stefanv:
Can we use one of the new theme primitives from Sphinx Design / PDST?

jarrodmillman:
### Before
2024-02-22T11:06:10,532757632-08:00

### After
2024-02-22T11:06:19,028008965-08:00

PR #515, opened 2024-02-28
Add field list shortcode and partial

jarrodmillman:
See https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/lists.html and https://github.com/scientific-python/scientific-python.org/pull/576

### DIFF
--- a/assets/theme-css/lists.css
+++ b/assets/theme-css/lists.css
@@ -1,0 +1,18 @@
+dl.field-list {
+  display: grid;
+  grid-template-columns: fit-content(30%) auto;
+}
+
+dl.field-list > dt {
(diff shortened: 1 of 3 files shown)


jarrodmillman:
{{ $entries := slice
  (dict "term" "John" "def" 30)
  (dict "term" "Will" "def" 28)
  (dict "term" "Joey" "def" 24)
}}

jarrodmillman:
Use `layouts/partial/_elements` or `layouts/partial/elements`  for partials intended to be used by shortcodes and other partials (e.g., `layouts/partials/field-list.html`.

PR #522, opened 2024-03-03
Remove custom youtube shortcode

jarrodmillman:
See https://github.com/scientific-python/learn.scientific-python.org/pull/150 and https://github.com/scientific-python/scientific-python.org/pull/580


stefanv:
Built-in youtube shortcode gained a lot of functionality recently, so +1 on removing ours.

PR #529, opened 2024-03-06
Refactor cards and items into partials

jarrodmillman:


stefanv:
Looks good; same request as with the others: can we add a docstring to the partial?

jarrodmillman:
I am still not sure this is the right direction, but I am pretty close to checking that it works for all the use cases. There will be a fair bit of clean up and refactoring once this is finished and I will document things at that point. But I should be able to sort things out in the next week or two.

PR #530, opened 2024-03-06
Refactor grid into partial

jarrodmillman:


layouts/partials/_elements/grid.html
```diff
@@ -0,0 +1,26 @@
+{{- $outline := default "false" .outline -}}
```
stefanv, line 1: Is it possible to add a comment here, without generating blank lines in the output, which documents the input parameters expected by the element?

layouts/shortcodes/grid1.html
```diff
@@ -121,12 +106,4 @@
...
-{{- range $key, $d := $items -}}
-  {{- if eq $d.type "card" }}
-    {{- partial "_elements/card.html" $d -}}
-  {{- else }}
-    {{- partial "_elements/item.html" $d -}}
-  {{- end }}
-{{- end }}
-  </div>
-</div>
+{{- partial "_elements/grid.html" (merge .Params (dict "items" $items)) -}}
```
stefanv, line 109: Same question on `.Params` as with the other PR. May well be necessary, just checking.

jarrodmillman: We have two optional parameters `outline` and `columns`.

PR #537, opened 2024-03-08
Use column value for smallest screens

jarrodmillman:
Use number of columns specified for xs screens for screens smaller than xs. Before it was hard-coded to 1.


jarrodmillman:
### Before

2024-03-08T10:51:01,636853173-08:00

### After

PR #542, opened 2024-03-12
Fix processing of scss

jarrodmillman:
Fix https://github.com/numpy/numpy.org/pull/732


stefanv:
Could you explain what this change does?

stefanv:
At some point we changed from iterating over filenames to iterating over resources. So, there's no need to get again.

stefanv:
xref: https://github.com/gohugoio/hugo/issues/12214

stefanv:
Details given in https://github.com/numpy/numpy.org/pull/732#issue-2176759475

PR #550, opened 2024-03-17
Add figure functionality for numpy.org

jarrodmillman:


assets/theme-css/figure.css
```diff
@@ -0,0 +1,43 @@
+img {
+  max-width: 100%;
+}
+
+img.align-left,
```
stefanv, line 5: Heh, on how many elements will we use this `align-left` class?!

assets/theme-css/figure.css
```diff
@@ -0,0 +1,43 @@
...
+figure.align-left,
```
stefanv, line 6: Do we parse this file with SCSS yet? If so, something like this might be slightly easier to read:

```css

#header,
#header-g {
    .l-header-top {
       height: 55px;
       line-height: 50px;
       border-top: 5px solid #474747;
       background-color: #f0f1f3;
       font-size: 16px;
    }
}
```

layouts/shortcodes/figure.html
```diff
@@ -7,7 +7,12 @@
 alt = 'Cute puppies'
 height = 200
 width = 200
-caption = 'A figure is an image with a caption and/or a legend:'
+title = 'Figure title'
+attr = '(**Figure Credits:** Daily cute puppy image from unslash.com)'
```
stefanv, line 11: Should we add the brackets programatically?

stefanv: We were thinking of another name... I think you suggested `credit` and `credit_link/url`? Otherwise `attribution` written out would work.

stefanv (commented):
Looks great, thanks!

layouts/shortcodes/figure.html
```diff
@@ -41,9 +48,15 @@
 {{- $m.SetInMap "image" "align" "center" -}}
```
stefanv, line 48: Why use scratch here, and why override alignment?

assets/theme-css/figure.css
jarrodmillman: I just moved this out of styles.css. The only new thing in this file is
```
strong.caption-title {
  font-weight: bold;
}
```
The rest came from PST dependencies, I believe. I will move it back for now and then investigate where it came from and move it to a different location to make it clear.

layouts/shortcodes/figure.html
jarrodmillman: Removed.

layouts/shortcodes/figure.html
jarrodmillman: I removed the example with parentheses and changed `attr` to `attribution`.

assets/theme-css/figure.css
jarrodmillman: We can refactor the old css to scss in a future PR.

assets/theme-css/figure.css
jarrodmillman: I removed the `figure.css` and added the new styling to `styles.css`.

jarrodmillman: We can look at `align-left` in a future PR.

PR #554, opened 2024-03-19
Center image in figures

jarrodmillman:
https://discourse.gohugo.io/t/is-it-possible-to-change-map-value-by-key/32538


stefanv:
Note to self: hugo does [not allow modification of dicts](https://discourse.gohugo.io/t/is-it-possible-to-change-map-value-by-key/32538) :eyes:

PR #561, opened 2024-03-23
Use rem for fonts, margins, and padding

jarrodmillman:
Use rem instead of px for font sizes, margins, and paddings. I left `assets/theme-css/navbar.css` and `assets/theme-css/hero.css` alone, since I plan to revisit them soon.

### DIFF
--- a/assets/theme-css/styles.css
+++ b/assets/theme-css/styles.css
@@ -109,7 +109,7 @@ a > code {
 
 @media only screen and (max-width: 600px) {
   .flex-column {
-    margin: 0 30px;
+    margin: 0 1.875rem;
   }
 }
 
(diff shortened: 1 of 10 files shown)


assets/theme-css/styles.css
```diff
@@ -130,13 +129,13 @@ svg.icon {
...
 .chroma {
-  padding: 10px;
+  padding: 0.625rem;
 }
 
 /* reboot stuff needed for pydata-sphinx-theme */
 
 pre {
-  font-size: 0.875em;
+  font-size: 0.875rem;
```
jarrodmillman, line 138: I am tempted to remove this. I prefer to keep the code the same size as the normal text.

stefanv: I presume this is to calibrate the regular font wise against code. They should perceptual ly be the same size. The adjustment here feels too large for that.

jarrodmillman: I removed the changes.

PR #572, opened 2024-03-27
Add tomlToTable

jarrodmillman:
Alternative to #548.

- [ ] refactor into a partial with a shortcode frontend


jarrodmillman:
https://gohugo.io/render-hooks/tables/

PR #573, opened 2024-03-27
Add link to scipy preview

jarrodmillman:


alphapapa:
@jarrodmillman 

-   It seems that the problem may have been misidentified.  The broken links appear to come from the `config.yaml` file in the scipy.org repo.  Those links start with a `/` so that they work when viewed from <http://scipy.org>.  They appear broken to the Netlify checker because it's running on a copy of the generated files loaded from a local directory.
    
    Jarrod linked to [Update hugo by jarrodmillman · Pull Request #729 · numpy/numpy.org · GitHub](https://github.com/numpy/numpy.org/pull/729), saying that a similar problem was fixed in it, but that appears to be different: the links changed were in content in Markdown files, not in the site header or general navigation links.

-   So I'm not sure if there is really a feasible solution to this problem.  The links need to stay as they are so they will work when served from the Web site.  I don't think it would be worth the trouble to post-process them just so the Netlify checker doesn't complain for this theme preview feature, which never gets published to a live site.

-   It seems like the best thing to do would be to exclude these previews from the link checker.  OTOH that checker could potentially catch some problems which these previews are intended to catch before going live.

- I found that we can use the `todoPatterns` in the `netlify.toml` file to mark those links as "todo", which prevents their appearing broken from causing the deploy preview to fail.  The docs aren't clear, so you can see how I tried 3 different pattern values before finding the ones that work.  Those commits can be squashed, of course.

Is this acceptable?  The only alternative I can think of would be to post-process the links in the local copies before running the link checker, which doesn't seem like a great idea.

Alternatively, if I have misidentified the problem myself, please point me in the right direction so I can apply a better fix.  Thanks.

jarrodmillman:
Err. This is a little more complicated than I thought and may have led to me figuring out some configuration issues on some of our other sites. I think just ignoring the links isn't the best option because it makes following links on the preview sites because when you look at the [scpiy preview](https://deploy-preview-573--scientific-python-hugo-theme.netlify.app/scipy/) the links don't work. For example, clicking on the `install` link yields [this](https://deploy-preview-573--scientific-python-hugo-theme.netlify.app/install) when it should be [this](https://deploy-preview-573--scientific-python-hugo-theme.netlify.app/scipy/install).

I wondered why other sites worked and found `relativeURLs: true` in `config.toml`. I am not sure why we set that, so we made need to investigate a bit. But I don't think we should set it. It seems to work ok when I remove it. See
- https://github.com/scientific-python/learn.scientific-python.org/pull/166
- #582

alphapapa:
@jarrodmillman Aha!  Good find of the `relativeURLs` option.  Turned out the solution was not to post-process the links but to have Hugo not pre-process them.  Thanks.

Issue #481, opened 2024-02-15
Kitchen sink -- generic items


jarrodmillman:
Start implementing this page in our kitchen sink:
https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/generic.html

Here is the rough order of priority:

- [ ] Math
  Hugo added support for math in markdown in v0.122.0 https://gohugo.io/content-management/mathematics/
  We should replace our custom implementation #366
  It looks like math is broken on the blog, so it would be good to get this done ASAP ([see this example](https://blog.scientific-python.org/networkx/atsp/understanding-the-ascent-method/))
- [ ] Keys / Menu labels
- [ ] Centered text
- [ ] Sidebars
- [ ] References

PR #589, opened 2024-04-23
New Hugo math equation support

alphapapa:
See https://github.com/scientific-python/scientific-python-hugo-theme/issues/481.


jarrodmillman:
I am a little confused about what Hugo actually added. It seems like we still have the same code, but we replace a regular expression to find the embedded math with several lines of configuration. I am not sure this is better than what we had. Sorry, I didn't look carefully at the description of the new feature.

I will take a closer look at this soon and see if it is much better than what we had before or what it gains us.

alphapapa:
Understood, will wait for further instructions.  Thanks.

jarrodmillman:
It looks like it is mainly adding the pass-through extension.
https://github.com/gohugoio/hugo-goldmark-extensions?tab=readme-ov-file#passthrough-extension
That is worth having I think.

stefanv:
I think using `$` as a delimiter is somewhat crucial, and looks like this configuration makes that hard to use: https://gohugo.io/content-management/mathematics/#inline-delimiters

PR #615, opened 2024-05-11
#607: Pagefind Search

brianhawthorne:
Add proof-of-concept pagefind site search.


Makefile
```diff
@@ -26,7 +26,10 @@ doc/content/shortcodes.md: $(wildcard layouts/shortcodes/*.html)
 # Serve for development purposes.
 serve-dev: doc-serve
 doc-serve: doc/content/shortcodes.md
-	(cd doc && hugo --printI18nWarnings serve --themesDir="../.." --disableFastRender --poll 1000ms)
+	(cd doc; \
+	hugo; \
+	yes | npx pagefind --site public; \
```
stefanv, line 31:
```suggestion
	yes | npx pagefind --site public; \
```

Maybe flipping yes and npx would be slightly more intuitive. Your call.

layouts/partials/search.html
```diff
@@ -0,0 +1,88 @@
+<link href="/pagefind/pagefind-ui.css" rel="stylesheet">
+<script src="/pagefind/pagefind-ui.js"></script>
+
+<style>
```
stefanv, line 4: Would be nice to have this defined outside of the template, in the assets/css dir.

layouts/partials/search.html
```diff
@@ -0,0 +1,88 @@
+<link href="/pagefind/pagefind-ui.css" rel="stylesheet">
+<script src="/pagefind/pagefind-ui.js"></script>
```
stefanv, line 2: Where are these files?

layouts/partials/search.html
```diff
@@ -0,0 +1,88 @@
...
+    backdrop-filter: blur(3px);
+   }
+  .search-dialog input {
+      padding: 10px 15px;
+      color: #333;
+  }
+  .pagefind-ui button {
+      border: none;
+  }
+  .pagefind-ui__result-title.svelte-4xnkmf .pagefind-ui__result-link.svelte-4xnkmf {
```
stefanv, line 27: Would be better if we didn't have to pull in svelte to make search work, but I can't examine JS yet to see how involved that would be.

stefanv (commented):
Thanks @brianhawthorne! This is looking good already.

Makefile
brianhawthorne: Agreed - I piped the `yes` command because I didn't initially see that npx supports a `--yes` flag; but it turned up buried on closer reading.

layouts/partials/search.html
brianhawthorne: Those files get generated when `pagefind` runs, so will not be part of this project's source code.  (https://pagefind.app/docs/)

layouts/partials/search.html
brianhawthorne: It's a nice feature of pagefind that it generates a reasonable search results listing for you out of the box.  The downside is if we use that default then we don't have much control over how it is rendered and styled, but I thought it was reasonable to lean on it at least for the proof of concept phase of this ticket.

I wasn't super happy about having to refer to these CSS classes here, especially the little hash-like suffixes, which don't convey an obvious meaning and I could imagine might change between pagefind releases.  An alternative would be to apply these two color unset rules (which make the results more legible against dark mode) by coding the selector to match the DOM structure instead of referring to these class names.  But that is also brittle since the generated DOM could also change shape between releases.

layouts/partials/search.html
```diff
@@ -0,0 +1,88 @@
...
+      color: unset;
+  }
+  input.pagefind-ui__search-input.svelte-e9gkc3 {
+      color: unset;
+  }
+</style>
+
+<dialog class="search-dialog"></dialog>
+
+<script>
```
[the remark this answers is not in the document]
brianhawthorne: I tried but wasn't able to resolve (404) the standalone js file.  I put it in `assets/js/search.js` with the URL "./js/search.js", following the pattern of CSS files under theme-css, but no luck.  Also tried adding and putting it under a dedicated `assets/theme-js` directory, so it would look more identical to the CSS case but that didn't work either.

So I'm inclined to leave the JS here unless we can figure out how to get hugo to serve it in a way that makes sense.  But it doesn't seem important enough to spend a lot of time on.

layouts/partials/search.html
brianhawthorne: Hmm, I added it under `assets/theme-css` instead of `doc/assets/css` since the search feature is provided by the theme and not just the example doc site.  That worked fine locally with the serve-dev make target, but it's blowing up in the CI build: https://app.netlify.com/sites/scientific-python-hugo-theme/deploys/66a68856fbc041000831976e

For now I'm reverting to put this CSS back where it was (inline) in the search partial, which worked, and affords the benefit of having (almost) the entire implementation in a single file, for easy delete or modify.

layouts/partials/search.html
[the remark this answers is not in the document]
brianhawthorne: Ok, I found the js bundling stuff, so I see no need to explicitly link to the js in the html as it will get swept into the bundle automatically.  Went ahead and separated it out into search.js.

layouts/partials/search.html
brianhawthorne: Figured out how to apply the color unset without referring to pagefind internal svelte classes.

brianhawthorne:
@jarrodmillman and @stefanv , all PR comments resolved, ready for another review

stefanv:
This looks great, @brianhawthorne. I see the search doesn't work on subpages, and will have to investigate why that is.

stefanv:
@brianhawthorne Do you mind enabling pushes to this branch (should be a little checkbox off to the right)?

brianhawthorne:
@stefanv , I can look into the problem with indexing subpages (good catch btw).  Do you have an example keyword that should pull up a subpage in a search but doesn't?

stefanv:
Ah, no, the problem is more obvious (on my build at least): the search dialog doesn't pop up when the button is clicked on subpages.

brianhawthorne:
@stefanv , this ought to get the search working on every page: https://github.com/scientific-python/scientific-python-hugo-theme/pull/615/commits/49fea620affb7b1a4ee7c18b8d90b3fd1a5081a4

Let me know if anything else seems amiss.

stefanv:
Strange Netlify error today:

```
11:26:32 AM: Started restoring cached Node.js version
11:26:32 AM: Finished restoring cached Node.js version
11:30:54 AM: Failed during stage 'Install dependencies': dependency_installation script returned non-zero exit code: 1
11:30:54 AM: Version '20' not found - try `nvm ls-remote` to browse available versions.
11:30:54 AM: Failed to install Node.js version '20'
11:30:54 AM: Failing build: Failed to install dependencies
```

brianhawthorne:
Ugh, yeah saw that yesterday.  Apparently having trouble installing node for one of the preliminary checks.  Haven't had a chance to look deeply into it yet, but I think it's probably unrelated to the changes in this PR, since it was building fine before my last commit, which just moved a small chunk of template from one Hugo .html file to another.

I can take a stab at debugging.  Was hoping it would be something you guys had seen before :/

stefanv:
I agree, I think this is on Hugo's side; I'll retry the CI again, now that they've had a day to fix it.

stefanv:
OK, latest rebuild worked; it won't report here, but it looks fine.

stefanv:
https://66eb407c558e39a4e02926d6--scientific-python-hugo-theme.netlify.app/

LGTM!

stefanv (approved):

PR #629, opened 2024-07-14
Use FontAwesome for backtotop icon

stefanv:
Some computers do not have the unicode fonts required to render ⮝


bsipocz (approved):
This works now 🎉

bsipocz:
Closes https://github.com/scientific-python/scientific-python.org/issues/658

PR #631, opened 2024-08-12
Allow specifying a canonical page URL

stefanv:


jarrodmillman:
Should we mention this somewhere in the documentation? Maybe somewhere here:
https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/doc/content/user_guide/features.md

stefanv:
I wasn't sure whether we wanted to encourage our site not being the canonical source. But I can add it to the docs, if you think it should be more widely visible.

PR #634, opened 2024-08-14
Update PDST assets

stefanv:


jarrodmillman:
I had been avoiding importing files we aren't using.

stefanv:
I can take those out; this was to test visual changes, to ensure we don't have too big a jump next time we do an ugprade.

stefanv:
Removed unused files. Applied different fix for missing variable, in response to https://github.com/pydata/pydata-sphinx-theme/issues/1953

assets/theme-css/pst/pydata-sphinx-theme.scss
```diff
@@ -1,3 +1,9 @@
+// Next line added by SPHT, to add missing bootstrap-provided variables
```
jarrodmillman, line 1: I've been adding things like this to the bottom of `assets/theme-css/styles.css`.

jarrodmillman: My plan was to factor `#pst-back-to-top` out of `base/_base.scss` and into something like `base/_back_to_top.scss` or something.

jarrodmillman: Why not just keep commenting out `  //  z-index: $zindex-tooltip;` like we are currently doing and then factor out the styling we don't need from `base/_base.scss`?

stefanv: I thought the less editing of the original sources needed, the easier it would be to stay in sync.

PR #640, opened 2024-09-09
Update hugo (0.134.3)

jarrodmillman:


stefanv:
Hrm, fails to find an image in the blog post, but that image is linked correctly.
Looks like it's trying to render that image on the index page, where the image isn't present: 

```
public/blog/index.html:416:14 <img src="checkin.png" alt="Morning group check-in">
```

stefanv:
xref: https://discourse.gohugo.io/t/potential-regression-between-0-132-and-0-134/51574?u=stefanv

stefanv:
Huzzah!

PR #647, opened 2024-10-14
Use `⌘` (`KeyboardEvent.metaKey`) with Pagefind search toggle on macOS

agriyakhetarpal:
## Description

This is a very small PR that should enable searching through the Command (⌘) key on macOS, which is the standard method to invoke the search bar followed by the PST. This allows both the use of <kbd>⌘</kbd> + <kbd>K</kbd> on macOS and the use of  <kbd>Ctrl</kbd> + <kbd>K</kbd> on Linux and Windows. I've added the `KeyboardEvent.metaKey` property for this based on the MDN docs: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey.

~~I don't think it is warranted to check for the host's operating system – I did think about it but since there is no reasonable alternative for this property on Windows-based keyboards besides the button for the Windows logo, this change is short and not too invasive in that sense. :)~~ I added some additional code to the event listener present so that we get the right tooltips based on the host's operating system. :)

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -51,7 +51,6 @@ window.addEventListener("DOMContentLoaded", () => {
     }
   });
 
-
   document
     .querySelector(".search-button")
     .addEventListener("click", showSearch);
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -50,8 +50,8 @@ window.addEventListener("DOMContentLoaded", () => {
 
   window.addEventListener("keydown", (evt) => {
     if (
-      evt.key === "k" &&
-      ((isMac && evt.metaKey) || (!isMac && evt.ctrlKey))
+      ((isMac && evt.metaKey) || (!isMac && evt.ctrlKey)) &&
+      evt.key === "k"
     ) {
       evt.preventDefault(); // prevents default browser behaviour
       toggleSearch();
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -45,10 +45,11 @@ window.addEventListener("DOMContentLoaded", () => {
   });
 
   window.addEventListener("keydown", (evt) => {
-    if (evt.key === "k" && evt.ctrlKey) {
+    if (evt.key === "k" && (evt.ctrlKey || evt.metaKey)) {
       toggleSearch();
     }
   });
+
 
   document
     .querySelector(".search-button")
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -2,9 +2,13 @@
 
 window.addEventListener("DOMContentLoaded", () => {
   let searchDialog = document.querySelector(".search-dialog");
+  let searchButton = document.getElementById("search-button");
 
   // Do nothing here if search is not enabled.
-  if (!searchDialog) return;
+  if (!searchDialog || !searchButton) return;
+
+  const isMac = /(Mac|iPhone|iPod|iPad)/i.test(navigator.userAgent);
+  searchButton.title = `Search (${isMac ? "⌘" : "Ctrl"} + K)`;
 
   new PagefindUI({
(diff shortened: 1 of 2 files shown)


layouts/partials/navbar.html
```diff
@@ -26,7 +26,7 @@
 
       <div class="navbar-end">
         {{- if .Site.Params.search }}
-        <button class="search-button" title="Search (Ctrl+k)">
+        <button class="search-button" title="Search (Ctrl/Cmd+K)">
```
agriyakhetarpal, line 29: While this should work, to truly achieve an implementation where it displays just "Cmd/⌘" on macOS and "Ctrl" on other platforms, I wrote an event listener for this – however, I wish to confirm if that's okay to add before I get towards committing/pushing those changes

agriyakhetarpal:
<kbd>⌘</kbd> + <kbd>K</kbd> works on the deploy preview at https://deploy-preview-647--scientific-python-hugo-theme.netlify.app/ on my macOS device.

layouts/partials/navbar.html
agriyakhetarpal: Actually, it's quite a small change. I added it to the existing event listener, so it should be easy enough for a review!

assets/js/search.js
```diff
@@ -2,9 +2,13 @@
...
 window.addEventListener("DOMContentLoaded", () => {
   let searchDialog = document.querySelector(".search-dialog");
+  let searchButton = document.getElementById("search-button");
 
   // Do nothing here if search is not enabled.
-  if (!searchDialog) return;
+  if (!searchDialog || !searchButton) return;
+
+  const isMac = /(Mac|iPhone|iPod|iPad)/i.test(navigator.userAgent);
+  searchButton.title = `Search (${isMac ? "⌘" : "Ctrl"} + K)`;
```
agriyakhetarpal, line 11: This is from https://gist.github.com/DWboutin/7cf15061157f2daf04ef1e25dd110b6a. I don't see any licensing information in there, but that could be omitted since it's a [widely used](https://github.com/search?q=%2F%28Mac%7CiPhone%7CiPod%7CiPad%29%2Fi.test&type=code) snippet.

assets/js/search.js
```diff
@@ -45,7 +49,11 @@ window.addEventListener("DOMContentLoaded", () => {
   });
 
   window.addEventListener("keydown", (evt) => {
-    if (evt.key === "k" && evt.ctrlKey) {
+    if (
+      evt.key === "k" &&
+      ((isMac && evt.metaKey) || (!isMac && evt.ctrlKey))
```
stefanv, line 54: I know this feels like a silly tweak, but if you'll humor me: could we swap these two lines around to mimic the order in which the keystroke is presented to the user?

stefanv (approved):
Minor suggestion, but also OK to merge without. Thank you @agriyakhetarpal!

assets/js/search.js
agriyakhetarpal: Haha, makes sense to me! Addressed in c2de9fd666257229963a68f13ba773b834ca1da4, apologies for the force push.

agriyakhetarpal:
Hi, @stefanv! I hope it is okay to ping you again here? Full context: I'm trying to use this functionality downstream in https://github.com/pybamm-team/pybamm.org/pull/210, which is where I noticed the problem and brought it up here – it would be great if this PR could be re-reviewed (and merged) soon! :)

We usually point the theme submodule to the latest commit, so I have no issues with waiting for a new release with this change, the PR would be enough (I'll point to a release when it is up and ready here).

agriyakhetarpal:
I see, a type label needs to be added (this might have been a new CI addition since it wasn't there last week).

PR #653, opened 2024-11-19
Add basic example site

jarrodmillman:


exampleSite/config.yaml
```diff
@@ -0,0 +1,60 @@
...
+  fonts:
+    - name: "Lato"
+      weights: [400, 900]
+  hero:
+    title: Example Site
+    image: logo.svg
+    buttontext: Get Started
+    buttonlink: "#"
+  navbar:
+    - title: User Guide
```
stefanv, line 38: Maybe another link here? Or at least one that resolves to something meaningful.

doc/content/user_guide/getstarted.md
```diff
@@ -42,7 +42,7 @@ This theme is designed to be used as a Git submodule inside your site's source r
 3. Copy the theme's documentation site as a template for your site:
```
stefanv, line 42:
```suggestion
3. Copy the theme's example site as a template:
```

exampleSite/config.yaml
jarrodmillman: We can clean this up later. For now, I just want to make sure we aren't asking folks to copy a ton of stuff they need to delete.

PR #669, opened 2025-03-07
feat: added hover effect on the clear button

Bashamega:
I have added a hover effect on the clear button.
image

### DIFF
--- a/assets/theme-css/search.css
+++ b/assets/theme-css/search.css
@@ -24,3 +24,10 @@
 .pagefind-ui__result-title > .pagefind-ui__result-link {
   color: unset !important;
 }
+.pagefind-ui__search-clear{
+  transition: background-color 0.3s ease, color 0.3s ease;
+}
+.pagefind-ui__search-clear:hover{
+  background-color: #333;
+  color: #fff;
+}


Bashamega:
I have opened a pr for the search button hover effect: 
https://github.com/scientific-python/scientific-python-hugo-theme/pull/669

stefanv:
Not sure why, the classes look OK, but it doesn't work for me.

Bashamega:
> Not sure why, the classes look OK, but it doesn't work for me in Firefox.

Weird. I don't know how to set up the project, so I have written the code in the inspec. Can you help me set it up?
When I run hugo server I get:
image
Thank you

Bashamega:
> Not sure why, the classes look OK, but it doesn't work for me in Firefox.

Hello:)
It is also not taking effect on my side. 
Yesterday I had to use the inspect tool to test the changes. Do I need to build it and how can I test the site locally? 
I couldn't find a guide to help me set up the theme for development.
Thank you very much

Bashamega:
Hello @stefanv,

Thank you for your reply.

This is my first time contributing to this repository or any Hugo theme. I wasn’t able to see my changes locally, so I copied my CSS changes and applied them via the browser's inspect tool.

Could you guide me on how to set up the theme locally?

Thanks!

agriyakhetarpal:
I think the reason why it doesn't work is because Svelte appends an identifier at the end of the class, so it's always `pagefind-ui__search-clear svelte-e9gkc3` or so on, so it would need to be accessed in the CSS as `.pagefind-ui__search-clear.svelte-e9gkc3`. I imagine we can't do that as this will be different on each build of the site.

Perhaps https://pagefind.app/docs/ui-usage/#customising-the-styles would be useful? It mentions:

> Pagefind UI can be styled manually by omitting the `/pagefind/pagefind-ui.css` stylesheet. In this case it will function as a pure HTML component.

so we could probably copy the styles and override the ones we need. However, that means we won't get any updates to the styles that Pagefind makes.

Issue #477, opened 2024-02-13
Accessibility meta issue


jarrodmillman:
**CAUTION: This is a rough placeholder for upcoming accessibility improvements and needs to be fleshed out in more detail.**

We leverage (and plan to leverage more) on the accessibility work being done for the PST (pydata sphinx theme). For example, see https://github.com/pydata/pydata-sphinx-theme/pull/1564

@trallard Here is a partial list of SPHT (scientific python hugo theme) specific work:

- [ ] Theme specific colors
  -  https://github.com/scipy/scipy/issues/19958
  -  https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/assets/theme-css/vars.css
- [ ] Improved accessibility automated testing
  -  https://github.com/pydata/pydata-sphinx-theme/blob/main/tests/test_a11y.py
- [ ] Buttons
  -  https://pydata-sphinx-theme--1564.org.readthedocs.build/en/1564/user_guide/web-components.html#badges-and-button-links
  - Unlike PST we will implement buttons as `<button>` and not `<a>` tags

smeragoel:
I have a question regarding the custom shortcodes, specifically the `yamltotables` component.

1. I don't see the actual component in the link provided (same for `youtube`)
2. I'll get more context when I see the component, but there's an [existing PST tables component](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/tables.html) and we are doing [active accessibility work](https://github.com/Quansight-Labs/czi-scientific-python-mgmt/issues/50) on it, so we can leverage this component if we aren't already doing so 😸

jarrodmillman:
@smeragoel Sorry about that. I removed the yamlToTable documentation (#570) because we are switching from yaml to toml for shortcodes (#484). There is a PR in progress, but I haven't finished it (#572).  I updated the link above to point to the PR and removed the deadlink. I am planning to use the PST styling in the new PR, but haven't yet made much progress other than adding to class tags used by PST.

I will try to get the tomlToTable PR ready to merge. Thanks for looking into this!

smeragoel:
Thanks for the update, @jarrodmillman! I'm happy to take a look at the table design once it's ready, so feel free to ping me.

[...]

PR #675, opened 2025-05-14
[DRAFT]: Enhancements for dropdowns: animations, open at initial state, and chevron indicators

agriyakhetarpal:
## Description

> [!TIP]
> At the moment, this PR is stemmed from and contains changes from #674. Please use this link for a better diff:

> [!NOTE]
> This PR is a draft in its current state, and is not yet ready for review.

This PR modifies the `dropdown` shortcode to add three new features, in order to establish feature parity with the dropdowns from the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#dropdowns) and the [sphinx-design](https://sphinx-design.readthedocs.io/en/pydata-theme/dropdowns.html) projects. This was also discussed in https://github.com/scientific-python/scientific-python-hugo-theme/issues/477#issuecomment-2037694735, and should resolve that part of the conversation.

In particular, the features added here are as follows:
- An "open" dropdown: the option to leave the dropdown open at initial load, i.e., with the contents expanded instead of collapsed
- Chevron indicators: these are now displayed at the end of the dropdown's title container when the dropdown has been expanded or is in its collapsed state, and are animated, and there is also a choice between
- Animations for the chevron indicators and for how the content is displayed (fade-in, fade-in-slide-down) have been added, based on the corresponding option for the directive

I've checked for any accessibility issues/implications, and I think this should be fine; I've closely followed the PST/SD implementation(s), and the chevrons here work similarly.


agriyakhetarpal:
I appreciate the feedback, @smeragoel! You are right, there is a slight delay between the darkening of the dropdown bar background on hover and the scaling of the chevrons kicking in. However, this is because the PST implementation in https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#dropdowns does not include animations; the dropdown bar's background changes colour and the chevrons do not animate when `transform: scale(1.1)` is applied, both of them instead scaling instantaneously.

I've gone ahead and implemented your suggestion with the above commit to match PST – the chevrons should no longer scale when hovering and thus end at the same time as the dropdown's background colour change. Could you PTAL when it is convenient for you?

<hr>

To be completely honest: I did notice this difference when working on the changes here, and I've quite come to like the slight deviation that I've made from PST – I feel that it's nice that the chevrons here also animate on hovering rather than abruptly scaling in size without an animation. However, I understand that I would be biased here as I'm familiar with my PR more :)

Could you educate me about any accessibility concerns that may arise with the previous case (i.e., before this commit) as I may not be aware of them? Thank you!

stefanv:
Looks good to me, thanks @agriyakhetarpal!

Could we make the chevron animation go in the other direction? It feels to me like, if the content scrolls *down* the chevron should rotate *clockwise*. Just my aesthetic feeling :)

agriyakhetarpal:
> Looks good to me, thanks @agriyakhetarpal!
> 
> Could we make the chevron animation go in the other direction? It feels to me like, if the content scrolls _down_ the chevron should rotate _clockwise_. This is already true for the sideways chevron, but not the up-down one.
> 
> Just my aesthetic feeling :)

Yes, certainly, I agree. This is also the case with the GitHub PR checks we can see above the comment box, they rotate clockwise in a 180-degree rotation and I like them too :)

agriyakhetarpal:
Actually, wait, the `down-up` chevron does rotate clockwise from down to up, doesn't it? Do you mean it should rotate clockwise from up to down as well, i.e., to complete a full-circle rotation in one close/open of a dropdown, or would you like it if it were to first rotate _anti_-clockwise from down to up + then clockwise from up to down (like as the GitHub PR checks one above)?

I ask this because I have no strong opinion here; the last example in https://sphinx-design.readthedocs.io/en/pydata-theme/dropdowns.html#more-examples is the down-up case, and it does not have an animation at all; I chose to deviate from it and add an animation as I felt that it was better to have one. :)

stefanv:
The GitHub directions feel intuitive to me, so I'd mimic that.

agriyakhetarpal:
One thing we should also address here before merging: the FontAwesome icons can take their due time to load on throttled networks based on the limited simulations I performed and on what I noticed with several hard refreshes (without caches) – the dropdown bars will show up first and the chevrons way later on a slow network. Maybe it makes sense to vendor these specific icons? I think PST is doing this, too.

stefanv:
Yes, either that, or you may be able to use unicode chevrons. They should look good enough.

assets/theme-css/sphinx-design/_dropdown.scss
```diff
@@ -5,10 +5,35 @@ details.sd-dropdown {
     font-weight: 700;
     // don't overlap the chevron
     padding-right: 3em !important;
-    -moz-user-select: none;
-    -ms-user-select: none;
-    -webkit-user-select: none;
```
agriyakhetarpal, line 10: My text editor warns that these attributes are non-standard and should be avoided, and `user-select` is baseline already. The same warnings occur in a few other places below, and I removed the browser-engine-specific styling there as well.

assets/theme-css/sphinx-design/_dropdown.scss
```diff
@@ -79,6 +103,61 @@ details.sd-dropdown {
...
+    transform: scale(1.1);
+  }
+
+  &[open] > .sd-summary-title:hover .sd-summary-chevron-right i {
+    transform: rotate(90deg) scale(1.1);
+  }
+
+  &[open] > .sd-summary-title:hover .sd-summary-chevron-down i {
+    transform: rotate(-180deg) scale(1.1);
+  }
```
agriyakhetarpal, line 159: In the earlier commit I pushed today, I changed this to `.sd-summary-title:hover` as we want to increase the chevron sizes only when hovering over the dropdown bars and not when hovering over the content inside them (this is the same as PST).

In particular, these lines address the cases where you hover over the chevrons and click on them – we don't want the chevrons to transition from a 1.1x scale to a 1.0x scale while rotating and then back to a 1.1x scale, but rather stay at a 1.1x scale throughout the transition.

assets/theme-css/sphinx-design/_dropdown.scss
```diff
@@ -90,18 +169,10 @@ details.sd-dropdown {
 
   // Transition animation
   &.sd-fade-in[open] summary ~ * {
-    -moz-animation: sd-fade-in 0.5s ease-in-out;
-    -webkit-animation: sd-fade-in 0.5s ease-in-out;
```
agriyakhetarpal, line 94: Same as above – these are non-standard attributes and are best avoided. I can also clean these up from everywhere in a separate PR and reduce the diff here if you want me to.

layouts/shortcodes/dropdown.html
```diff
@@ -31,23 +31,152 @@
...
+    <span class="sd-summary-text">
+    {{- with $data.title }}
+      {{- . }}
+    {{- else }}
+      <i class="fas fa-ellipsis-h no-title sd-octicon"></i>
+    {{- end }}
+    </span>
+    <span class="sd-summary-state-marker {{ $chevronClass }}">
+      <i class="fas {{ $chevronIcon }}"></i>
+    </span>
```
agriyakhetarpal, line 179: `.sd-summary-state-marker` is a separate element that makes it easier for us to style where the chevrons are placed (at the end of the dropdown bars).

agriyakhetarpal (commented):
All done (for now). This is ready for another look! Here is a self-review for some explanations.

layouts/shortcodes/dropdown.html
stefanv: `sd-` belongs to Sphinx Design, so is this prefix intentional?

assets/theme-css/sphinx-design/_dropdown.scss
```diff
@@ -5,10 +5,35 @@ details.sd-dropdown {
...
+      align-items: center;
+      justify-content: center;
+      width: 0.75em;
+    }
+
+    // The hover effect should only change opacity, not transform.
+    // We are transforming the chevron elements instead, see below.
+    &:hover .sd-summary-state-marker svg {
+      opacity: 1;
+    }
```
agriyakhetarpal:
```suggestion
    &:hover .dropdown-summary-state-marker svg {
      opacity: 1;
    }
```

stefanv:
I think we either make the directions coherent across components, or remove the animation on the up-down variant, or remove the up-down variant entirely.

Issue #618, opened 2024-06-06
BUG: dark mode issues


bsipocz:
https://github.com/scientific-python/scientific-python.org/pull/645#issuecomment-2150514389

PR #679, opened 2025-11-19
Do not adjust logos for viewing in dark mode

agriyakhetarpal:
The code originating from the PST stylesheets here: https://github.com/scientific-python/scientific-python-hugo-theme/blob/762065ef3bfafd8a091d2df8d079dac4fbc0b2c8/assets/theme-css/pst/variables/_color.scss#L305-L320 unfortunately causes the Scientific Python logo and any custom logos provided by users to appear darker and with increased contrast in dark mode. This PR addresses that issue by adding the `dark-light` class to all logos to ensure compliance with that rule. I view this as a bug and as an unintended side effect of using these stylesheets, as I believe it would be better for us to leave the decision to adjust logos in dark mode to the respective projects through their own stylesheets.

This is another one of the current dark mode issues; xref #618. Another piece of reasoning behind this: our dark mode support is already pretty sensitive, and we don't want projects' logos to inadvertently break WCAG 1.4.3 when switching to dark mode if their colours are very close to the minimum-contrast threshold!

P.S. I thought about removing this code section from the stylesheets I linked above entirely, but since it's coming from the PST, I decided not to, assuming there must be a reason they included it, and deviating from the PST styles in general warrants a separate discussion elsewhere.

<hr>

#### BEFORE
image

#### AFTER
image

### DIFF
--- a/assets/theme-css/footer.css
+++ b/assets/theme-css/footer.css
@@ -39,6 +39,18 @@
   color: var(--spht-color-link-hover);
 }
 
+.footer-item a.nav-external {
+  white-space: nowrap;
+}
+
+.footer-item a.nav-external::after {
+  font: var(--fa-font-solid);
+  content: "\f35d";
+  font-size: 0.75em;
+  margin-left: 0.3em;
+  display: inline-block;
+}
+
 .footer-actions {
   max-width: 25vw;
   margin: 1.25rem 1.563rem;
--- a/assets/theme-css/navbar.css
+++ b/assets/theme-css/navbar.css
@@ -41,6 +41,15 @@ a.navbar-item:hover,
   text-decoration-thickness: max(3px, 0.1875rem, 0.12em);
 }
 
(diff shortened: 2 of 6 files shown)


stefanv:
Should we not perhaps avoid adjusting image contrast overall?

agriyakhetarpal:
Likely, yes. I don't know the basis for adding that CSS rule to the PST was, though. It would indeed be nice not to adjust image contrast at all. However, I'm still inclined to discuss this separately, as it comes from an external stylesheet, and it would be better for us that if we deviate from the PST's stylesheets, we document it somehow...

Also, this is currently less of a breaking change than expected right now (it caught no one else's eye, for one :P), and part of that is also because none of the websites (besides ours at PyBaMM) have had dark mode enabled by default anyway.

agriyakhetarpal:
For now, yes! I'll open another issue about removing these dark mode image contrast modifications in general.

PR #685, opened 2025-12-04
Highlight more of Scientific Python's social media in our footer

agriyakhetarpal:
@stefanv, I saw these lying around in the repo; I wonder if it makes sense to add them? We're active on Bluesky, but not on TikTok and Instagram. I'm not sure if TikTok is available in the U.S., either. But I suppose it couldn't hurt to keep these for an international audience.

### DIFF
--- a/doc/config.yaml
+++ b/doc/config.yaml
@@ -56,8 +56,6 @@ params:
     socialmedia:
       - link: https://github.com/scientific-python/
         icon: github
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Done!

stefanv (approved):
People are still somewhat allergic to X it seems :)

PR #693, opened 2026-01-01
Add support for grid gutters to control spacing between grid items

### LINKED ISSUE #596: Improved grid functionality
- [x] https://sphinxdocs.ansys.com/version/stable/examples/sphinx-design.html#grid-card-columns #603 #604
- [ ] https://sphinxdocs.ansys.com/version/stable/examples/sphinx-design.html#grid-gutter

agriyakhetarpal:
This PR closes #596. This was already supported in our sphinx-design CSS, just that we didn't expose it as a parameter in the `grid` shortcode, so it was _partially_ implemented (no pun intended :P).

The spacing scale formula is defined here:

https://github.com/scientific-python/scientific-python-hugo-theme/blob/b3056c3a6a9b71813641860e1817aab4086ab55e/assets/theme-css/sphinx-design/_spacing.scss#L1-L9

which generates the responsive `sd-g-*` classes that control the `--sd-gutter-x` and `--sd-gutter-y` variables.

The gutter parameter can either take one or four integers between 0 to 5. The default is 2 for backwards compatibility. The four integers in the value, if set, correspond to the `xs`/`sm`/`md`/`lg` breakpoints. See the Bootstrap docs for more, which is what sphinx-design relies on: https://getbootstrap.com/docs/5.2/layout/grid/#grid-options

### DIFF
--- a/layouts/partials/_elements/grid.html
+++ b/layouts/partials/_elements/grid.html
@@ -19,7 +19,9 @@
   {{- $gSm := $gXs -}}
   {{- $gMd := $gXs -}}
   {{- $gLg := $gXs -}}
-  {{- if gt (len $gutterParts) 1 -}}
+  {{- if and (gt (len $gutterParts) 1) (ne (len $gutterParts) 4) -}}
+    {{- errorf "grid gutter must be either 1 or 4 values (but got %d): %q" (len $gutterParts) $gutter -}}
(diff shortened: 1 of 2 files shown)


stefanv (commented):
Two small comments, but also OK to just merge this directly.

stefanv (approved):

PR #706, opened 2026-02-26
Add Lato as a default font

agriyakhetarpal:
the `fonts:` section in the Hugo config is supposed to be for _additional_ fonts that users want to include for whatever reason, but we've been relying on Lato for downstream websites and it makes sense to add it here. Also enables Noto Sans which was disabled for some reason (inadvertently, based on the incorrect comment?).

Closes #705 

- [x] check Blog site deploy preview, which doesn't have Lato yet until https://github.com/scientific-python/blog.scientific-python.org/pull/268 will be merged
- [x] check Learn site deploy preview, which doesn't have Lato
- [x] check SciPy

### DIFF
--- a/assets/theme-css/navbar.css
+++ b/assets/theme-css/navbar.css
@@ -18,6 +18,7 @@
 .navbar-logo-text {
   font-size: 1.25rem;
   font-weight: 700;
+  font-family: "Lato", sans-serif;
 }
 .navbar .container {
   min-height: 4rem;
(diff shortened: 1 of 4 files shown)


stefanv:
I think only Scientific Python sites need Lato, so not sure it should be the *default* fallback. But it *should* be specified as an additional font in all SP websites.

agriyakhetarpal:
Hmm, in that case, we should close this PR and add Lato to all the websites' config. I don't feel very strongly about this PR; I just thought it's a good way to introduce consistency for all websites. Closing, then!

PR #712, opened 2026-07-23
Change .Site.LanguageCode to .Site.Language.Locale

kreczko:
This PR fixes #711 by replacing the deprecated `.Site.LanguageCode` to with `.Site.Language.Locale` (see [hugo docs](https://gohugo.io/methods/site/language/#locale))

PR #715, opened 2026-08-11
First version of replacing hugo-atom-feed (issue #714)

### LINKED ISSUE #714: Updating to Hugo 0.156.0 and later: `site.Author` broken, what to do with Atom feeds, and so on
As part of the discussion in #692, we stumbled into a problem that will arise when updating to Hugo 0.156.0 or later versions released to date, with the latest version at the time of writing being 0.164.0.

Currently, we use https://github.com/kaushalmodi/hugo-atom-feed as a theme for our theme, to add Atom feed functionality. It has been unmaintained for years, as unfortunate as that situation is. Hugo 0.124.0 deprecated the `site.Author` field, and it was fully removed with Hugo 0.156.0. Currently, there is an open PR to fix this: https://github.com/kaushalmodi/hugo-atom-feed/pull/22, but it is unclear to us when it will be merged (if ever).

There are some issues with all approaches through which we can think of tackling the situation:

- Vendoring https://github.com/kaushalmodi/hugo-atom-feed: the code there has always been GPL-licensed. Vendoring `layouts/_default/list.atom.xml`, as I was unknowingly trying in my earlier iterations of #692, would mean we can no longer use the BSD license, which we do not want.

- Drop the submodule and think about reintroducing Atom feeds on another day, and only rely on Hugo's built-in RSS for now: this would work well, but we don't have a sense of how many people use them on the current sites using the theme, I reckon. It would be nice to find out if this breaks any of our users. As far as I was able to find, only https://blog.scientific-python.org enables Atom feeds.

- On the possibility of migrating to the two alternative Atom feed options I posted in https://github.com/scientific-python/scientific-python-hugo-theme/pull/692#issuecomment-4317480648:
    - https://github.com/yelosan/hugo-feeds **cannot** be used, because it has its own licensing mess. There is GPL code in the repo by Kaushal Modi and other authors from 2018 to 2020, during which `list.atom.xml` was developed on. The license was swapped from GPL to MIT in 2022 during their Codeberg migration 🤦🏻 We should not rely on it without clarification from its maintainers.
    - https://github.com/hugo-fixit/hugo-atom-feed is fine, but it was not designed for our theme; rather, it was designed for the FixIt theme. We'll need to fork it and fix the layout (which is a bit of work, and I am unsure if we want to go that far to fix the Atom feed). It covers roughly half of what we need.

- We can write our own template against [RFC 4287](https://www.rfc-editor.org/rfc/rfc4287), and use that as a clean room implementation of a permissively licensed Atom feed syndicator for Hugo themes (or at least our theme needs, at the very least. This can be minimal on purpose and does not need to depend on the semantics of https://github.com/kaushalmodi/hugo-atom-feed, which includes many details we have not needed. I know about the Atom format, but I have to admit I do not know enough about it yet to write my own feed integrator. Note that this is something we'll need to do entirely on our own, as opposed to using any sort of AI coding agent assistance, even minimal. This is because AI systems have likely embedded https://github.com/kaushalmodi/hugo-atom-feed in their training data, which would mean we would quickly run into copyright concerns regarding the fidelity of an implementation and clean-room-ness.

kreczko:
Fixes #714 

## RFC 4287 implementation of a Atom hugo template

The in-tree Atom template follows RFC 4287 by providing:

* Required feed metadata: `title`, `id`, `updated`, and `author`.
* Required entry metadata: `title`, `id`, and `updated`.
* Optional metadata: `subtitle`, `generator`, `published`, categories, language, HTML content, and `self`/`alternate` links.
* Valid empty feeds, stable timestamps, feed-size limits, and HTML feed discovery.

Tests cover XML validity, required element cardinality, metadata values, entry ordering and limits, RFC 3339 dates, HTML escaping, categories, author fallback, empty feeds, and discovery links.

## AI assistance

AI tools were used to:

* help interpret the requirements of RFC 4287;
* help design tests for specific RFC 4287 requirements and edge cases;
* assist with debugging template-rendering and test failures.
* to make my notes readable for this PR

The resulting implementation and generated output were reviewed and validated through the repository’s test suite.

## Other notes

This is intentionally a basic implementation, focused on the core requirements of RFC 4287 and the features currently used by the Scientific Python blog feed.

Additional behaviour such as per-entry authors, related-post links, tracking parameters, or multilingual links can be added in response to concrete use cases and review feedback.

Pixi was introduced to provide a reproducible, pinned Hugo and Dart Sass environment for these tests. It also avoids installation-specific behaviour such as Snap confinement (my dev machine) and could be reused by other theme tests and development tasks. It proved convenient here, but the change is not essential to the Atom implementation and can be removed if maintainers would prefer to keep the existing tooling.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
-[submodule "themes/hugo-atom-feed"]
-	path = themes/hugo-atom-feed
-	url = https://github.com/kaushalmodi/hugo-atom-feed
+[submodule "themes/hugo-atom-layout"]
+	path = themes/hugo-atom-layout
+	url = https://github.com/FAST-HEP/hugo-atom-layout
+	
--- a/config.yaml
+++ b/config.yaml
@@ -1,7 +1,7 @@
 disableKinds:
   - taxonomy
 theme:
-  - scientific-python-hugo-theme/themes/hugo-atom-feed
+  - scientific-python-hugo-theme/themes/hugo-atom-layout
 params:
   colorScheme: auto # can be auto (browser setting), light, or dark
   images:


agriyakhetarpal (changes requested):
Wow, thanks @kreczko! I think it is great to have our own RFC 4287 support. The Pixi integration is also nice, which we've discussed on a few occasions before, but I haven't gotten around to implementing it.

However, in my humble opinion, I don't believe this repository is the best place to introduce this code, as such. Rather, it would be better if this could be hosted on your GitHub account or elsewhere. We could then rely on your work by adding it as a submodule in the repository, just like the current means of integration of `hugo-atom-feed`. I think it will be fine to mention there that the Atom feed provided only works for features desired by the https://github.com/scientific-python/scientific-python-hugo-theme repository, and that you can consider adding more features provided there is interest and provided you are willing to maintain it going forward.

The reason is that this is a lot of code, and it will be something that will be tricky for us to maintain going forward, if we were to add it. I know this layout is unlikely to change often, but the premise of these extra tests and related test scaffolding to run just for the sake of Atom feed support does not give me confidence that we will be able to maintain them.

Similarly, adding Pixi and a lockfile is also desirable, but it needs to be its own PR instead, as you suggest.

kreczko:
> However, in my humble opinion, I don't believe this repository is the best place to introduce this code, as such. Rather, it would be better if this could be hosted on your GitHub account or elsewhere.

Understood. Since I am only using the scipy hugo template for FAST-HEP, would it be OK to move this code to https://github.com/FAST-HEP/ ?
Then this PR would only contain the submodule change.

>  The Pixi integration is also nice, which we've discussed on a few occasions before, but I haven't gotten around to implementing it.
> ...
> Similarly, adding Pixi and a lockfile is also desirable, but it needs to be its own PR instead, as you suggest.
If that's desirable, I can move that to a separate PR - do you have a link to the discussion or a summary as to what is expected?
Do you envision a partial or full replacement of the Makefiles? e.g.
```toml
[tasks.doc-build]
cmd = "hugo --themesDir=../.."
cwd = "doc"

[tasks.doc-serve]
cmd = "hugo --printI18nWarnings serve --themesDir=../.. --disableFastRender --poll 1000ms"
cwd = "doc"

```

Also, what about the change to `baseof.html`?
```
   {{- range .AlternativeOutputFormats }}
    <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink }}">
    {{- end }}
```
This is to advertise the additional output formats - it would be blank if they are not defined.

Finally, while working on this I came across some updates for newer Hugo versions:
[The _default folder is removed](https://gohugo.io/templates/new-templatesystem-overview)

Everything in `layouts/_default` needs to move to `layouts` - if I don't forget, I will create a new issue for this.

agriyakhetarpal:
> > However, in my humble opinion, I don't believe this repository is the best place to introduce this code, as such. Rather, it would be better if this could be hosted on your GitHub account or elsewhere.
> 
> Understood. Since I am only using the scipy hugo template for FAST-HEP, would it be OK to move this code to [@FAST-HEP](https://github.com/FAST-HEP/) ? Then this PR would only contain the submodule change.

Yes, that would be great; thank you!

> > The Pixi integration is also nice, which we've discussed on a few occasions before, but I haven't gotten around to implementing it.
> > ...
> > Similarly, adding Pixi and a lockfile is also desirable, but it needs to be its own PR instead, as you suggest.
> If that's desirable, I can move that to a separate PR - do you have a link to the discussion or a summary as to what is expected?
> Do you envision a partial or full replacement of the Makefiles? e.g.
> 
> ```toml
> [tasks.doc-build]
> cmd = "hugo --themesDir=../.."
> cwd = "doc"
> 
> [tasks.doc-serve]
> cmd = "hugo --printI18nWarnings serve --themesDir=../.. --disableFastRender --poll 1000ms"
> cwd = "doc"
> ```

I don't think we've ever gotten around to creating a discussion/summary issue, unfortunately. However, we are on the same page – we should fully replace the Makefiles, since the steps are all one-liners and are not complicated. We can probably share the same environment across all tasks with a shared Hugo installation, so we don't waste much space (which I think is the default behaviour?).

Going a little further, our overall discussion with Pixi integration was also to do so for the websites this theme caters for. This way, it would be easier to build them using commands via Pixi package definitions in their repositories. This way, it would make for a nicer developer experience.

> Also, what about the change to `baseof.html`?
> 
> ```
>    {{- range .AlternativeOutputFormats }}
>     <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink }}">
>     {{- end }}
> ```
> 
> This is to advertise the additional output formats - it would be blank if they are not defined.
> 
> Finally, while working on this I came across some updates for newer Hugo versions: [The _default folder is removed](https://gohugo.io/templates/new-templatesystem-overview)
> 
> Everything in `layouts/_default` needs to move to `layouts` - if I don't forget, I will create a new issue for this.

I'd say the change in `baseof.html` is useful to keep here, and so is moving the files in `layouts/_default` to `layouts`. I think it would be easier if we migrate the theme according to newer Hugo versions one-by-one – we can have a PR that adds your Atom feed integrator as a submodule and updates to 0.156.0, thereby fixing #714, and then we already have #712, so we can probably put the `layouts` change in there and update to Hugo 0.164.0.

kreczko:
I have moved the changes to https://github.com/FAST-HEP/hugo-atom-layout and reduced the PR to
- a submodule change
- config update (theme)
- Hugo version update to 1.58.0 (netlify)

Note: The `baseof.html` change is now in [layouts/partials/head.html](https://github.com/FAST-HEP/hugo-atom-layout/blob/main/layouts/partials/head.html) which scipy theme uses.

.gitmodules
```diff
@@ -1,3 +1,4 @@
-[submodule "themes/hugo-atom-feed"]
-	path = themes/hugo-atom-feed
-	url = https://github.com/kaushalmodi/hugo-atom-feed
+[submodule "themes/hugo-atom-layout"]
+	path = themes/hugo-atom-layout
+	url = https://github.com/FAST-HEP/hugo-atom-layout
+
```
agriyakhetarpal: Nit: no newline at the end of file

kreczko: indeed, it was a `tab` instead of a blank line

kreczko:
Note: I had to update to 0.158.0 for the compatibility of `site.Language.Locale` - the old `languageCode` is ignored.

agriyakhetarpal (commented):
Thank you, @kreczko! I am testing the Atom feed generated from the deployment preview of this PR here: https://deploy-preview-715--scientific-python-hugo-theme.netlify.app/blog/atom.xml, versus the one here: https://blog.scientific-python.org/atom.xml

I use NetNewsWire (https://netnewswire.com) as my feed reader. I wonder if you could discuss those with me here and answer my possibly stupid questions about the two differences I found, considering you have experience in this field:

1. The original Atom feed also adds a `?utm_source=atom_feed` to the links, while the Fast-HEP implementation does not. While I'm not usually a fan of tracking parameters, I think they were added for a reason, and we should remain compatible with what has existed before. Could we add it?
2. The original Atom feed adds a blockquote preamble for each blog post: Image, i.e., "mplhep is a Scikit-HEP package that turns matplotlib into [...]", while Fast-HEP's does not: 
	Image
    Could we make it grab the summary from each page and add it to the feed in the same way as before?

The rest of it looks all great now, and I can confirm that no output is lost anywhere – I compared quite a few blogs with text and images to check! I will be happy to approve this PR once these comments are resolved, and we can get this rolling 🚀

kreczko:
Thanks for the feedback.
- utm_source added in https://github.com/FAST-HEP/hugo-atom-layout/commit/4bd91da2daeecc8770c6d121c87ee3061231e7d3
- blockquote added in https://github.com/FAST-HEP/hugo-atom-layout/commit/e645ff1ee61f9c7a1b816b7a5b41e48556bc5f96

agriyakhetarpal (approved):
Awesome, thanks a lot for your work @kreczko!

@stefanv, could you please skim through the discussion in #714 and here, and if you're happy with this, can we land it and unblock new Hugo releases? I would have merged this, but I thought I should ask you too – in my view, https://github.com/FAST-HEP/hugo-atom-layout here is functionally the same as https://github.com/kaushalmodi/hugo-atom-feed. Even with the AI assistance that @kreczko sought to complete this work, the much-reduced scope of the Fast-HEP implementation versus the latter (as I found when comparing the depth of `list.atom.xml` for both) makes me personally have no qualms in stating that this is fine enough in terms of licensing and should not incur a violation of the GPL. There simply isn't a significant difference that can be made across implementations when it comes to RFC 4287.

This layout is also beneficial in the sense that it's more tailored to our use case, and considering that @kreczko and Fast-HEP are aligned towards using our theme for their project's website—and also open to contributing towards other parts of our project, as was the case with adding Pixi on earlier iterations of this PR—it alleviates my concerns about maintainability for the medium term (and we've never needed a lot of maintenance for the Atom side of things anyway – it is only very recently, when it started breaking).

stefanv:
@agriyakhetarpal If you have carefully examined the output (it looks like you have) and are satisfied with it, I approve this change; the simplicity and being able to land patches in the new subtheme is worth it.

Thank you for this contribution, @kreczko!

agriyakhetarpal:
Thanks @stefanv! Yes, I am satisfied with this change in its current state, and my experiments with this feed did not find anything striking. @kreczko, is any extra change needed here – say, anything that may have been added to https://github.com/FAST-HEP/hugo-atom-layout that may have evolved further since our last conversation, and needs to be brought here as well? Let's merge it whenever you confirm that we are good to go :D

kreczko:
@agriyakhetarpal nothing new, ready to merge :).

agriyakhetarpal:
Perfect, thanks a lot!

PR #718, opened 2026-09-11
Advertise Atom/RSS feeds from the HTML head

agriyakhetarpal:
This PR changes our layout to emit `<link rel="alternate">` tags for every alternative output format of the current page, plus the home page's formats on all other pages. This lets feed readers autodiscover the Atom feed from any page on a site that uses our theme.

This is a bug in my view, since https://github.com/FAST-HEP/hugo-atom-layout does this via its own `partials/head.html`, but our partial took precedence and shadowed that partial.

Note that those websites that don't disable RSS will now also advertise Hugo's default `index.xml`.


layouts/partials/meta.html
```diff
@@ -2,3 +2,12 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 <meta http-equiv="x-ua-compatible" content="ie=edge">
+{{/* Advertise feeds (Atom, RSS) so that feed readers can discover them from any page. */}}
+{{ range .AlternativeOutputFormats -}}
+<link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink }}" title="{{ $.Site.Title }}">
+{{ end -}}
+{{ if not .IsHome -}}
+{{ range site.Home.AlternativeOutputFormats -}}
```
stefanv, line 12: I guess we may see some duplication here if both the page and site settings provide alternatives?

agriyakhetarpal: I think we can do something like having the page-level link having the page title, and the site-wide link having the site title. I guess that would work?

See 037aa3d98c0781c4b2f6eb6e041862cba519a846

stefanv: Yes, that works.

stefanv: Can also do a join, like `site name — page name` or similar.

agriyakhetarpal: Right; good idea. Sphinx also does that. Let's do it. :)

agriyakhetarpal: b744232ec68733f3e3b7b2c8ea56cbe995a2ccb7

PR #720, opened 2026-09-16
Update hugo-atom-layout to limit the home feed to main sections

agriyakhetarpal:
My PR FAST-HEP/hugo-atom-layout#1 is now merged, so this PR updates the submodule to its `main`, which includes those changes. This allows the home feed to be limited to the main section, restoring previous behaviour, but also provides an escape hatch for websites that also want to include additional sections. We don't have any websites in the Scientific Python ecosystem that need such a hatch (yet), so I've deliberately kept it undocumented (in our docs) for now. It remains documented in the hugo-atom-layout repository, here: https://github.com/FAST-HEP/hugo-atom-layout#which-pages-appear-in-the-home-feed

I verified that the Atom feed for https://github.com/scientific-python/blog.scientific-python.org now works perfectly with this change, and no other configuration change is required. This PR, when merged and released into 0.23.1, should unblock https://github.com/scientific-python/blog.scientific-python.org/pull/274.


stefanv:
@agriyakhetarpal What isn't mentioned in that PR is that this also seems to be the lever to *switch off* updates on pages, or am I misreading? For most of my sites, I want the atom feed only for the blog + its tags.

stefanv (approved):

stefanv:
Thanks very much :)

agriyakhetarpal:
> @agriyakhetarpal What isn't mentioned in that PR is that this also seems to be the lever to _switch off_ updates on pages, or am I misreading? For most of my sites, I want the atom feed only for the blog + its tags.

I am not sure I fully understood this comment, @stefanv. It is more that two things are involved; (i) what feeds exist in the first place, and (ii) which entries the home feed (i.e., the top-level `atom.xml` field) contains.

The first one is controlled by `outputs`: https://gohugo.io/configuration/outputs and https://github.com/scientific-python/blog.scientific-python.org/blob/53515d78dcfddbb2d212d6e3ed555062bb4dbeea/config.yaml#L36-L38, which is why we see https://blog.scientific-python.org/atom.xml as the home feed, and have several (61) other fields per taxonomy term (https://blog.scientific-python.org/tags/numpy/atom.xml for NumPy, https://blog.scientific-python.org/tags/gpu/atom.xml for GPU, etc.).

We changed only (ii) with this PR. That is to say, we were including extra entries within the home field (for authors, about pages, etc.), but now we include only the pages from the main section. The main section has not been explicitly configured in the blog website configuration; rather, it is automatically picked up by Hugo as the section with the most entries (and for the blog website, that happens to be `posts`, with 84 entries: https://blog.scientific-python.org/posts/). The same used to happen with https://github.com/kaushalmodi/hugo-atom-feed.

So, if I did understand you correctly, I would say this option is not a way to switch off Atom feeds off for non-blog pages, as those never had any feeds of their own – it is just that those non-blog pages won't show up as entries in the main/home Atom feed.

